PR #67, opened 2025-06-03
fix(auth-guard): add health endpoint to auth-exceptions

Dentling:
In a kubernetes-setup the `health`-endpoint is used for the readiness- and startup-probes, to update a pod without downtime.

The health-endpoint should in my opinion therefore be excluded from the auth-guard.


Wellington01:
> In a kubernetes-setup the `health`-endpoint is used for the readiness- and startup-probes, to update a pod without downtime.
> 
> The health-endpoint should in my opinion therefore be excluded from the auth-guard.

@Dentling 

You are right,

But reviewing the pull request and how the health method is implemented today, it is not working correctly.

We need to think about how to adjust it, if you want to take a look it will be a pleasure.

Wellington01:
Close because de feature resolved in other PR

PR #235, opened 2025-09-10
fix gitlab host

never934:
Refactored the GitLab API client initialization to consistently use the host defined in the `API_GITLAB_HOST` environment variable. Previously, the host was conditionally set from the `gitlabAuthDetail` object. This change ensures a standardized GitLab host configuration for all API interactions.


Wellington01 (changes requested):
Hi @never934 , hope you’re doing well!

I was reviewing the change made in the PR and this is not going to work as expected. Currently, when a user has a self-hosted GitLab instance, they must use a token during registration. In this scenario, the UI displays a “self-hosted” option along with a field for entering the GitLab self-hosted URL. I attached a screenshot to show how the front end looks in this case. This information is saved to the database, which is why the code was structured that way.

```
decrypt(gitlabAuthDetail.accessToken),
            ...(gitlabAuthDetail.host && { host: gitlabAuthDetail.host }),
```

Let me understand your issue more clearly: Is this happening during the registration of a new account? If so, I debugged the process and discovered that checking the “self-hosted” option currently breaks the app. I’ll prioritize getting this fixed so the flow works as intended.

If you’re encountering a different problem, please send me the details so I can investigate further. Sound good?

Screenshot 2025-09-10 at 17 16 47

Wellington01:
@never934 Hey, take a look at the permissions — since it connected successfully, something’s probably missing.

Issue #357, opened 2025-11-07
Add Code Review Preview mode to test configuration changes


malinosqui:
### Description

Create a “Code Review Preview” feature that allows users to test their current configuration (rules, prompts, etc.) and see how Kody would behave in a real Pull Request.

This should help users validate their setup before enabling it across their repositories.

### How it works

- Add a “Generate Preview” button in the configuration page.
- The user selects a closed PR to simulate a review.
- The system runs the configured pipeline (rules, prompts, etc.) and shows the results as if it were a normal review.
- Display a summary of the preview: suggestions found, files analyzed, and example comments.

### Acceptance criteria

- User can trigger a preview from the config screen.
- User can select a closed PR to run the preview.
- The review pipeline executes normally and shows suggestions based on current configuration.
- No comments are actually posted to the PR (simulation only).
- Handle error states gracefully (e.g., missing permissions, invalid PR).

### Mock
https://drink-polo-05683069.figma.site/

PR #381, opened 2025-11-10
all plans can now use BYOK

jairo-litman:
This pull request modifies the permission validation service to allow all plan types to utilize Bring Your Own Key (BYOK) configurations.

Previously, `MANAGED` plan types were explicitly restricted to using Kodus's managed keys. This change removes that hardcoded restriction, enabling the system to retrieve and apply a BYOK configuration for any plan type if one is available.

The updated logic ensures that:
*   A BYOK configuration is always attempted to be retrieved.
*   If a plan type explicitly requires BYOK (e.g., Free/BYOK plans) and no BYOK configuration is found, an error is thrown.
*   Otherwise, if a BYOK configuration is present, it will be used, regardless of the plan type.

### DIFF
--- a/src/ee/shared/services/permissionValidation.service.ts
+++ b/src/ee/shared/services/permissionValidation.service.ts
@@ -142,16 +142,16 @@ export class PermissionValidationService {
                 validation.planType,
             );
 
-            let byokConfig: BYOKConfig | null = null;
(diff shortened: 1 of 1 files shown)


malinosqui:
@kodus -v business-logic https://github.com/kodustech/kodus-ai/issues/357

PR #416, opened 2025-12-02
feat: implement SSO support for Google Workspace (fixes #413)

### LINKED ISSUE #413: Add basic SSO support for Google Workspace (OIDC)
### Description  
We need to introduce initial SSO support for enterprise customers using Google Workspace as their identity provider. This first version should work alongside our existing Auto Join feature, without replacing or breaking the current flow.

The goal is to support authentication via OIDC for organizations that request it while keeping the logic for workspace routing and auto membership exactly as it is today.

#### SSO  
- Allow an organization to enable SSO using Google Workspace  
- Store minimal SSO configuration for the org (issuer, client ID, client secret, redirect URLs)  
- When a user enters an email belonging to the org’s domain, trigger the SSO login flow  
- After successful authentication, create or link the user account  
- Support both mixed mode (SSO optional) and SSO enforced (only SSO allowed) in the future, but not required now

#### Auto Join coexistence  
SSO should not replace Auto Join. Both must work together following this logic:

1. User enters an email  
2. We identify the organization via domain  
3. If the org has SSO enabled  
   - Redirect user to SSO for authentication  
4. After successful SSO authentication  
   - If Auto Join is enabled for that domain  
     - Automatically add the user to the workspace  
   - If Auto Join is disabled  
     - Require invite or approval as today  
5. If the org does not have SSO enabled  
   - Keep the current login flow unchanged  
   - Auto Join continues working exactly as today

This ensures that:
- SSO handles **authentication**  
- Auto Join handles **workspace membership**  
- Both can be active without conflicting

### Acceptance criteria  
- Users from organizations with SSO enabled are able to authenticate via Google Workspace  
- Users from organizations without SSO continue with the normal login flow  
- Auto Join still works exactly the same after SSO authentication  
- A user authenticating with SSO enters the correct workspace if Auto Join is enabled  
- No breaking changes to existing login and signup flows  
- Feature is disabled by default and only enabled for selected organizations

soulcloude00:
This pull request implements Single Sign-On (SSO) support for Google Workspace, allowing users to authenticate via their organization's Google credentials. It includes the necessary backend logic for initiating the SSO flow, handling the OAuth callback, and managing user provisioning and organization membership.

**Key Changes:**

*   **SSO Verification (`CheckSSOUseCase`):**
    *   Added a use case to verify if a provided email address belongs to a domain configured for SSO.
    *   Generates the Google OAuth 2.0 authorization URL with the appropriate client ID, redirect URI, and state (containing the organization ID) if SSO is enabled.
*   **SSO Login & Callback Handling (`SSOLoginUseCase`):**
    *   Implemented the logic to handle the OAuth callback, exchanging the authorization code for access and ID tokens via Google's API.
    *   Verifies the ID token to retrieve user details.
    *   **JIT Provisioning:** Automatically creates a new user account if one does not exist for the authenticated email.
    *   **Organization Membership:** Ensures the user is added as a member of the organization associated with the SSO configuration.
    *   **Auto-Join Logic:** Automatically adds the user to other organizations that match their email domain.
    *   Generates and returns application authentication tokens (access/refresh tokens) upon successful processing.
*   **API Endpoints:**
    *   `POST /auth/sso/check`: Endpoint to check SSO availability for an email and retrieve the redirect URL.
    *   `GET /auth/sso/callback`: Endpoint to handle the Google OAuth redirect.
*   **Configuration:**
    *   Defined `SSOConfig` type (Client ID, Secret, Issuer, Domains, etc.).
    *   Added `SSO_CONFIG` to `OrganizationParametersKey` enum to support storing SSO settings in organization parameters.
*   **Module Updates:**
    *   Updated `AuthModule` to register the new use cases and import `OrganizationParametersModule`.
    *   Updated `OrganizationModule` to export `GetOrganizationsByDomainUseCase` for use in the SSO login flow.
    * Fixes #413


malinosqui:
Hey @soulcloude00, thanks for the PR! This was already addressed by @jairo-litman, though.

PR #687, opened 2026-02-17
let kody reply in pr reviews

### LINKED ISSUE #631: Kody didn't answer my questions
Kody didn't answer my questions. I tried sending the standard messages from the doctor, but nothing new was sent. She responds to new messages, but not to the new questions I asked.

Image

Image

Image

Sahil-Gupta584:
closes #631

This pull request enhances Kody's ability to participate in pull request review conversations by allowing it to respond to comments that are replies to its own previous comments, even if Kody is not explicitly mentioned in the reply.

Key changes include:
*   **Improved Conversation Detection**: Kody now identifies comments as part of a `CONVERSATION` if they are replies to existing comments on GitHub, Bitbucket, or Azure Repos, even if no specific command is recognized.
*   **Contextual Reply Handling**: The logic for ignoring comments has been updated. Kody will no longer ignore a comment if it is a direct reply to a comment previously made by Kody, enabling it to follow up on its own feedback or questions.
*   **Cross-Platform Parent Comment Identification**: A new utility method was introduced to accurately determine the parent comment across different Git platforms (GitHub, GitLab, Bitbucket, Azure Repos), which is essential for the updated reply detection logic.
*   **Expanded Webhook Triggering**: GitHub webhooks will now trigger Kody's chat functionality for any comment that is a reply, ensuring these conversational interactions are processed.

These changes allow Kody to maintain context and engage in more natural, multi-turn conversations within pull request reviews.

### DIFF
--- a/libs/platform/application/use-cases/codeManagement/chatWithKodyFromGit.use-case.ts
+++ b/libs/platform/application/use-cases/codeManagement/chatWithKodyFromGit.use-case.ts
@@ -644,6 +644,10 @@ export class ChatWithKodyFromGitUseCase {
                 },
             });
 
+        const commentsMap = new Map<any, Comment>(
+            allComments?.map((c) => [c.id, c]) || [],
+        );
+
         const commentId = this.getCommentId(params);
         const comment =
             params.platformType !== PlatformType.AZURE_REPOS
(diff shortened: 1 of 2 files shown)


Wellington01:
Actually, following up on the previous review: we realized there is simply no reliable way to make this work right now. 

Because we support Personal Access Tokens (PATs) and CLI integrations, Kody's comments are often posted via real user accounts. Since we don't have a guaranteed watermark on every single comment type or a database table tracking every comment ID Kody creates, there is no foolproof way to know if a parent comment was actually written by the AI or by a human.

Because of this structural limitation, we cannot implement this feature right now. It would behave unpredictably for users using PATs/CLI. We really appreciate the effort you put into this PR, but we'll have to hold off on this approach. Thank you anyway!

PR #730, opened 2026-02-27
Add E2B IP tunneling via Shadowsocks transparent proxy

malinosqui:
Clients with IP-restricted GitLab/GitHub can now use E2B sandboxes by routing traffic through our AWS fixed IP via a Shadowsocks proxy baked into a custom E2B template.

- Add e2b-sandbox/ with template definition, build script, and example configs
- Add createSandbox() with template support and fallback to default sandbox
- Skip apt-get install when using pre-built template (git/ripgrep/proxy included)
- Add E2B_TEMPLATE_ID env var; gitignore sensitive proxy config files

---

This pull request introduces E2B IP tunneling capabilities by integrating a Shadowsocks transparent proxy into the E2B sandbox environment.

Key changes include:

*   **E2B Sandbox Template (`kodus-sandbox`)**: A new E2B template is defined that includes `git`, `ripgrep`, `iptables`, and the `shadowsocks-rust` client (`sslocal`). This template is configured to automatically set up a transparent proxy.
*   **Shadowsocks Transparent Proxy**: `iptables` rules are configured within the sandbox to redirect all outbound TCP traffic through the `sslocal` client. This ensures that all network requests originating from the sandbox are routed via a Shadowsocks server, providing IP tunneling.
*   **Optimized Sandbox Creation**: The `E2BSandboxService` is updated to prioritize the use of a pre-built E2B template (specified by `E2B_TEMPLATE_ID`). When a template is used, `git` and `ripgrep` are no longer installed at runtime, improving sandbox startup performance.
*   **Fallback Mechanism**: If a specified E2B template fails to create or is not configured, the service gracefully falls back to creating a default E2B sandbox and installing dependencies on the fly.
*   **Configuration Management**: Example configuration files (`config.example.json` for Shadowsocks and `iptables-rules.example.sh`) are provided, and the actual configuration files (`config.json`, `iptables-rules.sh`) are added to `.gitignore` to prevent sensitive information from being committed.

### DIFF
--- a/scripts/dev/fetch-env-prod.sh
+++ b/scripts/dev/fetch-env-prod.sh
@@ -161,6 +161,7 @@ KEYS=(
     "/prod/kodus-orchestrator/API_MORPHLLM_API_KEY"
 
     "/prod/kodus-orchestrator/API_E2B_KEY"
+    "/prod/kodus-orchestrator/API_E2B_TEMPLATE_ID"
 
     "/prod/kodus-orchestrator/API_BETTERSTACK_API_TOKEN"
     "/prod/kodus-orchestrator/API_BETTERSTACK_HEARTBEAT_ERROR_RATE_URL"
--- a/scripts/dev/fetch-env-qa.sh
+++ b/scripts/dev/fetch-env-qa.sh
@@ -121,6 +121,7 @@ KEYS=(
     "/qa/kodus-orchestrator/API_MORPHLLM_API_KEY"
 
     "/qa/kodus-orchestrator/API_E2B_KEY"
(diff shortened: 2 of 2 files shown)


.env.example
```diff
@@ -247,6 +247,7 @@ WEBHOOK_FAILURE_WINDOW_MINUTES=30
 
 # E2B
 API_E2B_KEY=
+E2B_TEMPLATE_ID=
```
Wellington01: @kody consegue criar uma Kody Rule, para garantir que as novas chaves em env sigam padrão de API_*, WEB_* e quando for global GLOBAL_*

PR #936, opened 2026-04-16
Fix/ignore files field

sartorijr92:
This pull request introduces the following changes:

*   **Improved User Experience for Ignore Paths Field**: Prevents the "ignore files" textarea in the code review settings from resetting its content while a user is actively typing. This is achieved by introducing an `isEditing` state that pauses external updates to the field's value when it is focused.
*   **Enhanced Configuration Persistence Logic**: Ensures that code review configuration changes are correctly saved directly to the database when centralized configuration is disabled. Previously, the system might have incorrectly attempted to create a pull request for centralized configuration even when it was not enabled.
    *   An explicit check has been added to verify if a centralized repository is enabled before attempting any centralized configuration operations, allowing the system to fall back to direct database storage if it's disabled.
    *   The logic for handling centralized pull requests has been refined to explicitly check for `mode === 'centralized-pr'` before short-circuiting the database save, ensuring that direct database saves are correctly processed when intended.
    *   New test cases have been added to validate the correct configuration saving behavior when centralized configuration is disabled.

### DIFF
--- a/apps/web/src/app/(app)/settings/code-review/[repositoryId]/general/_components/ignore-paths.tsx
+++ b/apps/web/src/app/(app)/settings/code-review/[repositoryId]/general/_components/ignore-paths.tsx
@@ -19,12 +19,13 @@ export const IgnorePaths = () => {
         [field.value],
     );
     const [draftValue, setDraftValue] = useState(fieldValue);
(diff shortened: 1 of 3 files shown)


sartorijr92:
@kody review

PR #938, opened 2026-04-16
Fixing invitation flow and rules of delete

sartorijr92:
This pull request introduces two main functional changes:

1.  **Enhanced User Deletion Logic**: The user deletion process has been updated from a soft delete (marking a user as `REMOVED`) to a hard delete. This change ensures that when a user is deleted, all associated data, including profile configurations, permissions, profiles, authentication records, and team memberships, are also explicitly removed from the database within a transaction, preventing orphaned data.
2.  **Improved Invitation Flow**: The team member invitation process has been refined to use the email of the user who initiated the invitation as the sender for the invitation email. Previously, the invitation email always originated from the organization's owner. Now, if the inviter's email is available, it will be used; otherwise, it will fall back to the organization owner's email. This provides a more accurate and personalized invitation experience.


sartorijr92:
@kody review

PR #941, opened 2026-04-17
fix(queue): add worker health guard, channel logging, and fix reconne…

Wellington01:
…ct config

Root cause: consumer_timeout (2h) closes the AMQP channel but amqp-connection-manager only watches the connection — so the channel stays dead and the worker becomes a zombie with no consumers.

Changes:

1. WorkerHealthGuardService (new)
   - Startup assertion: waits for all managed channels to connect, exits if any fail within 30s so ECS can replace.
   - Channel-level zombie detector: listens for channel close/error events. When a channel dies while the connection is alive, forces reconnect. If channels stay dead for 90s, triggers graceful shutdown via SIGTERM (preserves shutdown hooks like releaseAllByInstance).

2. RabbitMQConnectionLoggerService (improved)
   - Adds close/error/connect listeners on every managed channel.
   - Logs instanceId, consumer count, and zombieRisk flag on every channel event for production observability.

3. rabbitmq.module.ts (config fix)
   - Moves reconnectTimeInSeconds inside connectionManagerOptions where amqp-connection-manager actually reads it (was silently ignored at top level since golevelup v9 upgrade).
   - Aligns client heartbeat to 60s to match the server config.

---

This pull request introduces significant improvements to RabbitMQ worker reliability, observability, and configuration.

Key changes include:

1.  **RabbitMQ Worker Health Guard**: Implements a new `WorkerHealthGuardService` to prevent "zombie worker" scenarios where RabbitMQ channels silently die while the connection remains active, causing workers to stop processing messages. This service:
    *   **Asserts channel connectivity on startup**: Ensures all configured channels connect successfully within a timeout, exiting if not.
    *   **Monitors channel events**: Listens for unexpected channel `close` and `error` events, attempting to force a connection reconnect when a channel dies but the main connection is still alive.
    *   **Performs periodic health checks**: Regularly verifies the state of the RabbitMQ connection and channels, initiating a graceful shutdown if the connection is lost for too long or if channels remain dead despite reconnect attempts.
    *   Includes a graceful shutdown mechanism to ensure proper resource release (e.g., inbox locks) before the worker exits.

2.  **Enhanced RabbitMQ Channel Logging**: Extends the `RabbitMQConnectionLoggerService` to provide more detailed insights into channel lifecycle events.
    *   Adds logging for channel `connect`, `close`, and `error` events.
    *   Includes the number of active consumers on channel connect and a `zombieRisk` flag on channel close events to highlight potential issues.

3.  **Refined RabbitMQ Connection Configuration**:
    *   **Corrects reconnect timing**: Moves `reconnectTimeInSeconds` to the `connectionManagerOptions` to ensure the intended 10-second reconnect interval is correctly applied, fixing a previous misconfiguration that caused it to default to 5 seconds.
    *   **Adjusts heartbeat interval**: Changes `heartbeatIntervalInSeconds` from 120 to 60 seconds for potentially faster detection of connection issues.
    *   **Adds new channels**: Introduces new RabbitMQ channel configurations (`channel-ast-graph-build` and `channel-ast-graph-incremental`) with specific prefetch counts, indicating new or optimized message processing capabilities.

### DIFF
--- a/apps/worker/src/worker-health-guard.service.ts
+++ b/apps/worker/src/worker-health-guard.service.ts
@@ -1,0 +1,444 @@
+import * as os from 'os';
+import { AmqpConnection } from '@golevelup/nestjs-rabbitmq';
+import { createLogger } from '@kodus/flow';
+import {
+    Injectable,
+    OnApplicationBootstrap,
+    OnApplicationShutdown,
+    Optional,
+} from '@nestjs/common';
+
+/**
+ * Guarantees that a running worker always has active RabbitMQ consumers.
+ *
+ * ## Why this exists
+ *
+ * amqp-connection-manager has a critical gap: when the broker closes a
+ * **channel** (e.g. consumer_timeout → PRECONDITION_FAILED), the
+ * ChannelWrapper sets `_channel = undefined` and waits for the
+ * **connection** to reconnect. But the connection is still alive, so
+ * nothing ever triggers channel re-creation. The consumer silently
+ * disappears and the worker becomes a zombie.
+ *
+ * ## How it works
+ *
+ * 1. **Startup assertion** — waits for all managed channels to connect.
+ *    If any channel fails within BOOT_TIMEOUT_MS, exits.
+ *
+ * 2. **Event-driven channel monitoring** — listens for `close` and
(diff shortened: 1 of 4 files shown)


Wellington01:
@kody start-review

PR #978, opened 2026-04-23
Allow user adding individual configs for many directories

### LINKED ISSUE #987: Allow users to select multiple directories when adding a configuration
Today, the system only lets the user select a single directory when creating a custom configuration. The idea now is to allow selecting multiple directories, treating them as a group that shares the same code review rules.

Also, the current rule for applying a directory configuration during a review is: it only applies if all PR files are داخل that selected directory. This will change — the new rule is: if at least one file from a configured directory is present in the PR, then that directory’s configuration should be used for the code review.

If a PR touches files from more than one configured directory group, the review should fall back to the repository-level configuration.

On top of that, we’ll add a `reviewScope` variable to the custom message. When used, it informs the user which configuration was applied during the PR review.

sartorijr92:
Closes #987 

This pull request introduces the capability to define code review configurations for **groups of multiple directories** within a single repository, enhancing flexibility beyond the previous limitation of one configuration per individual directory.

Key changes and their functional impact include:

*   **Directory Grouping:** Users can now create a single code review configuration that applies to several distinct file paths (folders) within a repository.
*   **Updated User Interface:**
    *   The "Add/Edit Repository Settings" modal has been redesigned to allow selecting multiple target directories for a single configuration.
    *   A new "Directory Group" component is introduced in the settings sidebar, displaying the primary folder path and indicating if additional folders are part of the same group.
    *   The Kody Rules section and other configuration displays now correctly reflect these directory groups, showing the primary path or group name.
    *   Dropdown menus for directory configurations now include an "Edit directories" option, allowing users to modify the folders within an existing group.
*   **Intelligent Configuration Resolution:**
    *   When a Pull Request (PR) is reviewed, the system now determines the most specific configuration based on affected files.
    *   If changed files in a PR fall within *exactly one* defined directory group, that group's specific configuration is applied.
    *   If changed files affect *multiple* directory groups, or no defined groups, the system gracefully falls back to the repository-level configuration.
    *   Files outside any configured directory group do not prevent a single-group match.
*   **Granular Management:** Users can now add or remove individual folders from an existing directory group. If a group is reduced to a single folder, it behaves as a single-folder group. Deleting the last folder in a group removes the entire group configuration.
*   **Enhanced Transparency:** A new `@reviewScope` placeholder is available for custom PR messages, providing clear information about which configuration level (global, repository, or specific directory group, including its primary path and count of other folders) was used for the review.
*   **Backward Compatibility:** Existing single-directory configurations are automatically treated as directory groups containing a single folder, ensuring no disruption to current setups.


sartorijr92:
@kody review

PR #1177, opened 2026-05-26
fix(agent-review): guard undefined agentResults before .map() access

### LINKED ISSUE #1176: fix(agent-review): guard missing agentResults before map
## Bug Description

`AgentReviewStage` can crash with:

```text
TypeError: Cannot read properties of undefined (reading 'map')
```

In our self-hosted Kodus deployment this surfaced as:

- `AgentReview::generalist` → `Generalist Agent — failed ... (TypeError): Cannot read properties of undefined (reading 'map')`
- `AgentReviewStage` → `partial_error` with the same message

## Observed Behavior

During a real PR review run, agent review progressed far enough to produce downstream review artifacts, but the stage still recorded a partial error because it tried to read `.map` on an undefined value.

This appears to happen when `result.agentResults` is unexpectedly undefined at runtime.

## Affected Code

`AgentReviewStage` assumes `result.agentResults` always exists in at least two places:

1. logging metadata:
- `libs/code-review/pipeline/stages/agent-review.stage.ts`
- around lines 437-451

```ts
agentResults: result.agentResults.map((r) => ({
```

2. collecting discarded suggestions:
- `libs/code-review/pipeline/stages/agent-review.stage.ts`
- around lines 549-566

```ts
for (const agentResult of result.agentResults) {
```

## Why This Is Surprising

The orchestrator contract says `agentResults` should always be an array:

- `libs/code-review/infrastructure/agents/review-orchestrator.service.ts`
- `OrchestratorOutput` defines `agentResults: ReviewAgentOutput[]`
- the orchestrator returns `agentResults: []` in the no-agent case and `agentResults` in the normal case

So either:

1. some runtime path is returning a malformed orchestrator result, or
2. `AgentReviewStage` is too trusting of a cross-layer contract and should defensively default to `[]`

## Reproduction Signal

This was observed on self-hosted review runs for a real Forgejo PR. The exact runtime row sequence included:

- `AgentReview::generalist` → error with `Cannot read properties of undefined (reading 'map')`
- `AgentReviewStage` → `partial_error` with the same message

## Expected Behavior

`AgentReviewStage` should not throw when `agentResults` is absent or malformed.

At minimum it should treat missing `agentResults` as `[]`, classify the failure normally, and continue using whatever suggestions/failures were already produced.

## Proposed Fix

Add a defensive default in `AgentReviewStage`:

```ts
const agentResults = result.agentResults ?? [];
```

and use that local variable for:

- metadata logging
- discarded suggestion aggregation

It may also be worth tracing the upstream call path to find why `agentResults` became undefined despite the orchestrator contract.

## References

- `libs/code-review/pipeline/stages/agent-review.stage.ts`
- `libs/code-review/infrastructure/agents/review-orchestrator.service.ts`

djdembeck:
## Summary

Guard two unguarded `result.agentResults` accesses in `AgentReviewStage` that crash with `TypeError: Cannot read properties of undefined (reading 'map')` when the orchestrator result omits the field.

Although `OrchestratorOutput` declares `agentResults: ReviewAgentOutput[]`, runtime paths exist where the field is `undefined` rather than an empty array (e.g. generalist agent dispatch failures). The stage currently trusts the contract unconditionally at:

- **line 443**: `result.agentResults.map(...)` — logging metadata
- **line 551**: `for (const agentResult of result.agentResults)` — collecting discarded suggestions

## Changes

- Default to empty array: `(result.agentResults ?? [])` at both call sites
- Add regression test for `agentResults: undefined` orchestrator result

Fixes #1176

### DIFF
--- a/libs/code-review/pipeline/stages/agent-review.stage.ts
+++ b/libs/code-review/pipeline/stages/agent-review.stage.ts
@@ -440,7 +440,7 @@ export class AgentReviewStage extends BasePipelineStage<CodeReviewPipelineContex
                 metadata: {
                     prNumber,
                     suggestionsCount: result.suggestions.length,
-                    agentResults: result.agentResults.map((r) => ({
+                    agentResults: (result.agentResults ?? []).map((r) => ({
                         agent: r.agentName,
                         category: r.agentCategory,
                         replicaIndex: r.agentReplicaIndex,
@@ -548,7 +548,7 @@ export class AgentReviewStage extends BasePipelineStage<CodeReviewPipelineContex
 
             // Collect suggestions discarded by severity filter and verify
             const allDiscarded: Partial<CodeSuggestion>[] = [];
-            for (const agentResult of result.agentResults) {
+            for (const agentResult of result.agentResults ?? []) {
                 if (agentResult.discardedBySeverity?.length) {
                     for (const s of agentResult.discardedBySeverity) {
                         allDiscarded.push({
--- a/test/unit/code-review/pipeline/stages/agent-review.stage.spec.ts
+++ b/test/unit/code-review/pipeline/stages/agent-review.stage.spec.ts
@@ -409,5 +409,46 @@ describe('AgentReviewStage', () => {
 
             expect(result.fileAnalysisResults).toEqual([]);
         });
+
+        it('should handle undefined agentResults without throwing', async () => {
+            mockOrchestrator.execute.mockResolvedValue({
+                suggestions: [
+                    {
+                        relevantFile: 'src/auth.ts',
+                        suggestionContent: 'Missing null check',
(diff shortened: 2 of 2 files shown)


djdembeck:
> kody code-review Kody Rules critical
> 
> The PR title starts with `fix(` but the description lacks a GitHub closing statement or issue reference, violating the '[Ensure feat/fix PRs close referenced issues](https://app.kodus.io/settings/code-review/global/kody-rules/18153886-f5a5-45bd-a3a3-c287086a5cf5?teamId=ebb24108-be7e-4fd9-9346-f1f2c7c0b7f1)' rule. Include a statement like 'Closes #123' or 'Fixes #456' in the description to properly link and track the resolution.

done

PR #1216, opened 2026-05-30
fix(setup): removing old integration when user click on update connection button

### LINKED ISSUE #1188: GitHub integration gets stuck after user revokes token - Update connection does not trigger re-auth
## Bug Description

When a user connects Kodus to GitHub via OAuth (GitHub App) and later revokes/deletes the token on GitHub's side, the integration enters a stuck state:

1. Kodus shows **"No repositories found"**
2. Clicking **"Update connection"** does NOT redirect to the GitHub authorization flow — it just reloads or navigates to the configuration page
3. The user cannot recover without manual database intervention

## Steps to Reproduce

1. Connect Kodus to GitHub via OAuth (Install GitHub App)
2. Go to GitHub → Settings → Applications → Authorized GitHub Apps
3. Revoke/delete the Kodus app authorization
4. Return to Kodus and try to select repositories → "Error saving repositories"
5. Click "Update connection" → stays on the same page, no redirect to GitHub

## Expected Behavior

When the GitHub token is invalid or revoked, clicking "Update connection" should trigger the OAuth flow again, allowing the user to re-authorize.

## Actual Behavior

The system detects `hasConnection = true` (because the integration record still exists in the database) and redirects to the configuration page instead of initiating OAuth. Since the token is invalid, `getRepositories()` returns an empty array, which the frontend interprets as "No repositories found".

### Frontend
In `apps/web/src/core/integrations/gitHubConnection.ts`:

```typescript
async connect(hasConnection: boolean, routerConfig: any, routerPath?: string) {
    if (hasConnection) {
        routerConfig.push(routerPath || `${routerConfig.pathname}/github/configuration`);
        return;
    }
    window.location.href = this.cfg.githubInstallUrl || "";
}
```

When `hasConnection = true`, the OAuth flow is never triggered. There is no check for whether the existing connection is actually valid.

### Backend
In `libs/platform/application/use-cases/codeManagement/get-repositories.ts` (line 95-107):

```typescript
catch (error) {
    this.logger.error({...});
    return [];  // Silently returns empty array on auth failure
}
```

When `instanceOctokit()` fails due to an invalid/revoked token, the error is caught and an empty array is returned. This gives no signal to the frontend that the connection is broken.

In `libs/platform/infrastructure/adapters/services/github/github.service.ts`:
- `getInstallationAuthentication()` (line 2205) tries to generate a token from a cached or stale `installationId`
- `getRepositories()` (line 1909) catches `BadRequestException` and throws it, which is then swallowed by the use case

## Suggested Fix

1. **Backend**: Return a distinct error code (e.g., `AUTH_INVALID` or `TOKEN_REVOKED`) when authentication fails, instead of returning an empty array or generic `BadRequestException`
2. **Frontend**: Detect the `AUTH_INVALID` error and force `hasConnection = false` behavior, triggering the OAuth redirect
3. **Alternative**: Add a "Reconnect" or "Re-authorize" button that bypasses the `hasConnection` check and always triggers OAuth

## Affected Users

At least 2 users reported this issue in support (see Discord threads). The workaround currently requires manual deletion of the integration record from the database.

## Related Files

- `apps/web/src/core/integrations/gitHubConnection.ts`
- `libs/platform/infrastructure/adapters/services/github/github.service.ts`
- `libs/platform/application/use-cases/codeManagement/get-repositories.ts`
- `apps/web/src/core/components/system/cardsGroup/index.tsx`

sartorijr92:
Closes #1188

### DIFF
--- a/apps/web/src/app/(setup)/setup/choosing-repositories/page.tsx
+++ b/apps/web/src/app/(setup)/setup/choosing-repositories/page.tsx
@@ -12,12 +12,16 @@ import { Page } from "@components/ui/page";
 import { toast } from "@components/ui/toaster/use-toast";
 import { ToggleGroup } from "@components/ui/toggle-group";
 import { useAsyncAction } from "@hooks/use-async-action";
-import { createOrUpdateRepositories } from "@services/codeManagement/fetch";
+import {
+    createOrUpdateRepositories,
+    deleteIntegration,
+} from "@services/codeManagement/fetch";
 import { useGetRepositories } from "@services/codeManagement/hooks";
 import {
     CODE_MANAGEMENT_API_PATHS,
     type Repository,
 } from "@services/codeManagement/types";
+import { INTEGRATION_CONFIG } from "@services/integrations/integrationConfig";
 import { fastSyncIDERules } from "@services/kodyRules/fetch";
 import { updateAutoLicenseAllowedUsers } from "@services/organizationParameters/fetch";
 import {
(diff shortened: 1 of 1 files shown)


sartorijr92:
@kody review --force

PR #1220, opened 2026-05-31
fix: skip fallback creation in writeAgentTrace to prevent orphaned stage logs

### LINKED ISSUE #1218: fix: writeAgentTrace race condition creates orphaned IN_PROGRESS stage logs
## Bug Description

In `agent-review.stage.ts`, the `writeAgentTrace()` method uses fire-and-forget for all agent progress events. When chunking occurs, `batch_started` fires **before** the recursive `execute()` call (which emits `started`), but both enter the microtask queue independently. If `batch_started`'s `findLatestStageLog` executes before `started` creates its record, it falls back to creating a new record — resulting in two `IN_PROGRESS` records for the same stage. One of them is never updated.

## Root Cause

Event ordering in chunked mode (`executeChunked()` in `base-code-review-agent.provider.ts`):

1. `batch_started` emitted (L1638) → `writeAgentTrace` (fire-and-forget)
2. `execute(batchInput)` called (L1663) → `started` emitted (L961) → `writeAgentTrace` (fire-and-forget)

Both `writeAgentTrace` calls are async with multiple `await` points. The race:

1. `batch_started`'s `findLatestStageLog` runs before `started`'s `updateCodeReview` creates the record → returns `null`
2. Fallback at L1881 creates a new `IN_PROGRESS` record for `batch_started`
3. `started` creates another `IN_PROGRESS` record via `updateCodeReview`
4. Subsequent events (`batch_completed`, `completed`) update only one of the two records
5. The other record stays `IN_PROGRESS` forever → UI shows "In Progress" permanently

## Evidence

Database shows 3 `AgentReview::generalist` records for a single execution:
- UUID 1893128d: permanently `in_progress`, message `"batch 1/2 starting"` — created by `batch_started` fallback, never updated

## Fix

In `writeAgentTrace()`, when `findLatestStageLog` returns `null` for non-`started` events, skip the fallback creation instead of creating a new record:

```typescript
// L1854-1888: Change the else branch
} else {
    const existing = executionUuid
        ? await this.automationExecutionService.findLatestStageLog(
              executionUuid,
              stageName,
          )
        : null;

    if (existing) {
        // ... update existing record (unchanged)
    } else {
        // Skip: don't create orphaned records when the started
        // event hasn't created the initial record yet.
        // The completed event will create the final record.
    }
}
```

This is the minimal fix — no architectural changes to the fire-and-forget pattern needed.

## Reproduction

Unit test confirming the race: `test/unit/code-review/pipeline/stages/agent-review.stage.spec.ts` — `writeAgentTrace - race condition` test suite.

The test fires `batch_started` with `findLatestStageLog` returning `null` (simulating the race), then fires `started`, and verifies that two `updateCodeReview` calls are made (documenting the current buggy behavior).

Fanduzi:
## Summary

- Remove fallback record creation in `writeAgentTrace()` when `findLatestStageLog` returns `null` for non-`started` events
- In chunked mode, `batch_started` fires before the recursive `execute()` emits `started` — both are fire-and-forget, creating a race where `batch_started`'s `findLatestStageLog` runs before `started` creates the initial record
- The fallback at L1879 then creates a second `IN_PROGRESS` record that never gets updated, leaving the UI stuck on "In Progress"

Closes #1218

## Test plan

- [x] `started` event still creates record via `updateCodeReview`
- [x] Non-started event with existing record updates via `updateStageLog`
- [x] Non-started event with no existing record skips fallback (no orphan)
- [x] Race condition simulation: `batch_started` then `started` creates only one record

```bash
yarn test --testPathPattern="agent-review.stage.spec"
```

### DIFF
--- a/libs/code-review/pipeline/stages/agent-review.stage.ts
+++ b/libs/code-review/pipeline/stages/agent-review.stage.ts
@@ -1876,15 +1876,29 @@ ${summaries}`,
                         existing.uuid,
                         updateData,
                     );
-                } else {
-                    // No existing record found — skip creating a new one.
-                    // In chunked mode, batch_started fires before the recursive
-                    // execute() emits started. Both are fire-and-forget, so
-                    // batch_started's findLatestStageLog may run before started
-                    // creates the initial record. Creating a fallback here would
-                    // produce an orphaned IN_PROGRESS record that never gets
-                    // updated. The completed event will create the final record.
+                } else if (
+                    status === AutomationStatus.SUCCESS ||
+                    status === AutomationStatus.ERROR
+                ) {
+                    // Fallback for terminal events (completed/error) that raced
+                    // ahead of 'started', or where 'started' failed to emit.
+                    // Without this, the final SUCCESS/ERROR state and agentTrace
+                    // metadata would be silently dropped.
+                    await this.automationExecutionService.updateCodeReview(
+                        filter,
+                        { status },
+                        label,
+                        stageName,
+                        metadata,
+                    );
                 }
+                // Non-terminal events (batch_started, investigating, etc.) with
+                // no existing record are silently skipped. In chunked mode,
+                // batch_started fires before the recursive execute() emits
+                // started — both are fire-and-forget, so batch_started's
+                // findLatestStageLog may run before started creates the initial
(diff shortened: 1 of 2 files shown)


Fanduzi:
Closing to resubmit with clean history.

PR #1236, opened 2026-06-02
fix(kody-rules): consistent orphan count, ACTIVE-only limit, no-data-loss updates, perf

Wellington01:


Wellington01:
@kody start-review

PR #1298, opened 2026-06-10
fix(boot): serialize prod migrations across replicas (advisory lock)

malinosqui:
## Problem

Self-hosted boots **api / worker / webhooks** from the same `docker/prod-entrypoint.sh` with `RUN_MIGRATIONS=true`, so they run the migration steps **concurrently** and race on `CREATE TYPE`:

```
error: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
Migration "Baseline..." failed ...
```

Under `set -e` that crashes the container → **CrashLoopBackOff churns the first boot/upgrade** (observed on a real droplet: api RestartCount=2, webhooks=2). And because the analytics-warehouse migration runs **after** the OLTP one in the same entrypoint, the warehouse is left uncreated during the window → `/cockpit/validate` **500**. The flaky `cockpit-analytics` e2e failure traces to exactly this race window.

It self-heals (a later restart wins the race), so it's not a hard outage — but it's a churny boot for every customer and it makes the analytics matrix coverage flaky.

## Fix

`scripts/run-migrations-with-lock.cjs` holds **one Postgres advisory lock** on the OLTP DB for the whole migration sequence (OLTP → ensure analytics schema → analytics warehouse). Concurrent replicas wait, then enter and find every migration already applied → fast no-op. The entrypoint calls it, with an **inline legacy fallback** if the script is somehow absent from the image, and the lock logic is **fail-open** (runs migrations anyway if the lock connection can't be established) — so this is never worse than today.

Two design points found while **testing locally against a real Postgres with two concurrent processes**:
- resolve the dist ormconfig from `process.cwd()`, not relative to `scripts/`;
- **poll with `pg_try_advisory_lock`**, not a blocking `pg_advisory_lock`: a blocked lock holds a snapshot for its whole wait, which stalls `CREATE INDEX CONCURRENTLY` in the replica that *is* migrating (CIC waits for all older snapshots) — a logical deadlock. Polling releases the snapshot between attempts.

## Validation

Local, two concurrent runners against a fresh DB:
- **before** the lock: 2nd runner failed with **3 duplicate-key errors, exit 1** (race reproduced).
- **after**: both exit **0**, **zero** duplicate-key — one runner migrates everything, the other waits then `No migrations are pending`.

## Still to validate before merge
- Full **multi-container image boot on a droplet** (script ships in image + entrypoint wiring + api/worker/webhooks together). The change is fail-open + has the inline fallback, but the image-boot integration hasn't been exercised yet. Recommend a droplet boot test (or let the next RC build exercise it) before merge.

## Related
- Pairs with #1297 (analytics fix forward-port). Once this lands, `cockpit-analytics` can be wired into the gating matrix reliably.

### DIFF
--- a/docker/prod-entrypoint.sh
+++ b/docker/prod-entrypoint.sh
@@ -52,27 +52,39 @@ RUN_MIGRATIONS="${RUN_MIGRATIONS:-false}"
 RUN_SEEDS="${RUN_SEEDS:-false}"
 
 if [ "$RUN_MIGRATIONS" = "true" ]; then
-  echo "▶ Running OLTP migrations (PROD)..."
-  # Use node with compiled migrations in prod
-  if [ -f "dist/libs/core/infrastructure/database/typeorm/ormconfig.js" ]; then
-      npm run migration:run:prod
+  # Run OLTP + analytics migrations under a Postgres advisory lock so the
+  # replicas that share this entrypoint (api / worker / webhooks) don't run
+  # them concurrently. Without serialization they race on CREATE TYPE
+  # (duplicate key pg_type_typname_nsp_index) -> migration fails -> set -e
+  # crashes the container -> CrashLoopBackOff churns the boot and the analytics
+  # warehouse (migrated after OLTP) is left uncreated -> /cockpit/validate 500.
+  if [ -f "scripts/run-migrations-with-lock.cjs" ]; then
+      echo "▶ Running migrations under advisory lock (PROD)..."
+      node -r ./tsconfig-paths-bootstrap.js scripts/run-migrations-with-lock.cjs
   else
-      echo "⚠️ Migration config not found at dist/libs/core/infrastructure/database/typeorm/ormconfig.js. Skipping."
-  fi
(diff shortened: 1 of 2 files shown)


malinosqui:
Closing in favour of a much simpler fix.

The migration race (concurrent CREATE TYPE -> `duplicate key pg_type_typname_nsp_index`) happens only because the installer's `.env` sets `RUN_MIGRATIONS=true` and **all** services (api/worker/webhooks/worker-analytics) inherit it via `env_file`, so they all migrate on boot.

The intended pattern is **single-migrator**, already used in-repo (`docker-compose.dev.yml` forces `RUN_MIGRATIONS=false` on mcp-manager because migrations "belong to the api/worker services, not here"). And cloud runs migrations **out-of-band** (`env-vars-generated.mdx`: set false for blue/green), so there's no per-replica race there.

So the advisory lock is unnecessary. Fix moves to kodus-installer: keep `RUN_MIGRATIONS=true` only on `api` (which runs OLTP **and** the analytics-warehouse migrations in the same entrypoint), set `false` on worker/webhooks/worker-analytics.

PR #1309, opened 2026-06-12
fix(e2e): resolve self-hosted API hostname so SSO + RBAC web scenarios pass

malinosqui:
## Why

The self-hosted release matrix has been failing on a single cell — `github × license-paid` — which blocks promotion (e.g. `selfhosted-2.1.18-rc.75` never became `selfhosted-2.1.18`). That cell fails on exactly four scenarios; the review engine itself passes everywhere:

- `rbac-frontend-routes`, `rbac-ui-render`
- `sso-cookie-domain`, `sso-multi-user`

All four share **one root cause**, not four flakes.

## Root cause

The harness hardcoded the API hostname as **`kodus-api`** (hyphen) for the web→API path and the SSO Caddy overlay. But the real installer container is **`kodus_api`** (underscore — `GLOBAL_API_CONTAINER_NAME`'s installer default, synced from this repo's `.env.example`/`.env.schema`) and the compose **service** is `api`. `kodus-api` resolves to *neither*:

- **RBAC**: the web's next-auth `authorize()` calls `http://kodus-api:3001/auth/login` → ENOTFOUND → returns `null` → the login **302s with no session**. Passes on cloud because there the web reaches the API over the public HTTPS URL, not a docker hostname — green-on-cloud / red-on-self-hosted was the tell.
- **SSO**: the overlay Caddy `reverse_proxy kodus-api:3001` never resolves → **"API never answered over Caddy internal-CA TLS"** at the TLS-wait step, before any assertion runs (hence "no [sso-e2e] FAIL line").

## Fix

Point every harness reference at the compose **service name `api`**, which Docker DNS always resolves on the shared network regardless of `container_name`:

- `tests/e2e/provisioning/self-hosted/vm.sh` — `WEB_HOSTNAME_API` → `api`; also **add `kodus-web` to the droplet log capture** (only `api/worker/webhooks` were captured, so the web-side 302 was invisible in evidence).
- `docker/sso-e2e/droplet/Caddyfile.tpl` + `Caddyfile.internal.tpl` — API upstream → `api:3001`.
- `scripts/sso-e2e/droplet/provision.sh` + `bootstrap-multi-user.sh` — `WEB_HOSTNAME_API` / repair → `api`.

The **local** lvh.me harness (`docker/sso-e2e/*`, `scripts/sso-e2e/run.sh`) builds its own api container literally named `kodus-api` and is intentionally left untouched.

## Risk

Low — test-harness only, cannot affect product or prod. `api` is the documented compose service name.

## Test plan

⚠️ **Not yet validated live** — these paths only run on a provisioned DigitalOcean droplet, not locally. Validated statically (YAML + `bash -n` + log cross-reference).

To validate without merging, dispatch the matrix against this branch on the github cell using the existing RC image:

```
gh workflow run e2e-self-hosted-matrix.yml --ref fix/e2e-selfhosted-api-hostname \
  -f image_tag=2.1.18-rc.75 -f only_provider=github
```

Caveat: this unblocks the API-over-TLS gate for SSO; the downstream Keycloak/SAML steps never ran in the failing matrix, so a later-stage failure could still surface that we haven't seen yet.

### DIFF
--- a/docker/sso-e2e/droplet/Caddyfile.internal.tpl
+++ b/docker/sso-e2e/droplet/Caddyfile.internal.tpl
@@ -40,7 +40,7 @@ app.${BASE} {
 
 api.${BASE} {
     tls internal
-    reverse_proxy kodus-api:3001 {
+    reverse_proxy api:3001 {
         import proxy_common
     }
 }
--- a/docker/sso-e2e/droplet/Caddyfile.tpl
+++ b/docker/sso-e2e/droplet/Caddyfile.tpl
@@ -46,7 +46,7 @@ app.${BASE} {
 }
 
 api.${BASE} {
-    reverse_proxy kodus-api:3001 {
+    reverse_proxy api:3001 {
         import proxy_common
     }
 }
--- a/scripts/sso-e2e/droplet/bootstrap-multi-user.sh
+++ b/scripts/sso-e2e/droplet/bootstrap-multi-user.sh
@@ -96,9 +96,9 @@ env_set() {
     fi
 }
 CURR_HOSTNAME=$(grep -E '^WEB_HOSTNAME_API=' /opt/kodus-installer/.env | head -1 | cut -d= -f2-)
-if [ "${CURR_HOSTNAME}" != "kodus-api" ]; then
-    echo "==> repairing WEB_HOSTNAME_API (was '${CURR_HOSTNAME}') -> kodus-api" >&2
-    env_set WEB_HOSTNAME_API "kodus-api"
+if [ "${CURR_HOSTNAME}" != "api" ]; then
+    echo "==> repairing WEB_HOSTNAME_API (was '${CURR_HOSTNAME}') -> api" >&2
+    env_set WEB_HOSTNAME_API "api"
     env_set WEB_PORT_API "3001"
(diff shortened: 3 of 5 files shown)


malinosqui:
✅ **Validated live on a real droplet** — dispatched against this branch on the github cell with the existing RC image:

```
gh workflow run e2e-self-hosted-matrix.yml --ref fix/e2e-selfhosted-api-hostname \
  -f image_tag=2.1.18-rc.75 -f only_provider=github -f matrix_file=matrix/full.yml
```

All four previously-failing scenarios now **PASS** ([run 27415769609](https://github.com/kodustech/kodus-ai/actions/runs/27415769609), conclusion: success):

| scenario | before | after |
|---|---|---|
| `sso-cookie-domain` | FAIL (API never answered over Caddy TLS) | **PASS** |
| `sso-multi-user` | FAIL (0/4) | **PASS** |
| `rbac-frontend-routes` | FAIL (302 no session) | **PASS** |
| `rbac-ui-render` | FAIL | **PASS** |

The downstream-SSO caveat in the description is resolved: `sso-cookie-domain` / `sso-multi-user` run the full Keycloak/SAML flow and went green end-to-end.

PR #1320, opened 2026-06-13
fix: openai base url

ALameLlama:
## Summary

I noticed I was getting this in code reviews

```
Generalist Agent — failed in 2085s (AI_RetryError): Failed after 2 attempts. Last error: Cannot connect to API: Headers Timeout Error

Generalist Agent — failed in 2085s (AI_RetryError): Failed after 2 attempts. Last error: Cannot connect to API: Headers Timeout Error
```


strubio-ray:
CI root cause from the failing jobs:

- Backend tests have a PR-caused failure in libs/code-review/infrastructure/agents/llm/env-llm-config.spec.ts. The changed env/model detection now returns openai_compatible with https://api.openai.com/v1 for the claude case, and returns configured=false for the Gemini/Vertex cases where the tests expected google_gemini/google_vertex. That needs either a code correction or updated expectations, depending on the intended provider-selection behavior.
- release-freeze eligibility is failing because the workflow token cannot post the status from this forked PR: Resource not accessible by integration (HTTP 403).
- Deploy web preview fails because Vercel secrets are unavailable on this forked PR, so the workflow runs vercel deploy with an empty --token value.
- Comment preview URLs also fails with Resource not accessible by integration while trying to create the preview comment.

So the only code-side issue visible here is the env-llm-config behavior/test mismatch; the other red checks are fork permission/secret limitations rather than PR code.

Wellington01:
Hey @ALameLlama — thanks for chasing this down, the timeout you hit is a real bug and you pointed us right at it. Going to close this one though, and here's the honest reasoning so it's not a black box.

The catch is that `openaiBaseURL` isn't just a value in this code, it's a signal. `isProxyBaseURL()` returns `true` for *any* base URL that isn't `api.anthropic.com`, and that flag drives provider selection in `env-llm-config.ts`:

- `isGemini && !viaProxy` → picks `google_gemini` / `google_vertex`
- `isClaude && openaiKey && !viaProxy` → picks `anthropic`
- otherwise → `openai_compatible`

By hoisting the `|| 'https://api.openai.com/v1'` default up to the `const openaiBaseURL = ...` declaration, the variable is now always set, so `viaProxy` is always `true`. That silently disables the native Gemini/Vertex/Anthropic paths and routes everything through `openai_compatible` at `api.openai.com` — which is exactly the two test failures in the CI run (claude → `openai_compatible`, gemini/vertex → `configured: false`). So it's a real behavior regression, not stale test expectations.

The actual gap your timeout came from is narrower — the one spot that passed a raw `undefined` straight into the SDK was `packages/kodus-common/src/llm/llmModelProvider.service.ts` (the `byok-to-vercel` sites already defaulted at the point of use). There's also a coupled change in `get-models-by-provider` (`/v1/models` → `/models`) that would break any operator whose `API_OPENAI_FORCE_BASE_URL` doesn't already end in `/v1`.

Rather than land a partial fix and chase the fallout, we're going to rework the base-URL resolution properly on our side so the "unset" signal and the SDK default stop being the same thing — and your timeout case gets fixed as part of it. I'll link the PR here when it's up so you can see it land. Really do appreciate you flagging it and digging into the CI.

ALameLlama:
@Wellington01 awesome sounds good, i was getting confused with some of the naming so a proper url resolved would be great

PR #1344, opened 2026-06-16
fix(ast-graph): make kodus-graph available in self-hosted worker

### LINKED ISSUE #1343: Self-hosted worker image does not expose kodus-graph to the runtime user
### Issue Summary

The self-hosted worker image installs `@kodus/kodus-graph`, but the binary is not available to the runtime user that starts the worker process.

In the production Dockerfile, `bun install -g @kodus/kodus-graph` runs before the image drops privileges with `USER 1000`. At runtime the worker runs as user `node` with `HOME=/home/node`, but the published worker image does not expose `kodus-graph` in `/home/node/.bun/bin` or in `PATH`.

`KodusGraphCli.install()` also checks for an existing binary with a raw `which kodus-graph`, while the parse/context commands already use `BUN_PATH_PREFIX` to include `$HOME/.bun/bin`.

Together, this makes the self-hosted AST graph runtime less reliable: the worker image appears to install `kodus-graph`, but the worker process may not be able to find it.

### Steps to Reproduce

1. Pull a published self-hosted worker image, for example `ghcr.io/kodustech/kodus-ai-worker:2.1.18`.
2. Run a shell as the image default runtime user.
3. Check the runtime user, home, PATH, `bun`, and `kodus-graph`.

Example:

```bash
docker run --rm --entrypoint sh ghcr.io/kodustech/kodus-ai-worker:2.1.18 -lc \
  'id; echo HOME=$HOME; echo PATH=$PATH; command -v bun || true; command -v kodus-graph || true'
```

### Actual Results

- The container runs as `uid=1000(node)`.
- `HOME=/home/node`.
- `bun` is available.
- `kodus-graph` is not found in the runtime user's PATH.

### Expected Results

- The self-hosted worker image exposes `kodus-graph` to the same non-root user that runs the worker.
- `KodusGraphCli.install()` checks for an existing binary using the same Bun global bin path prefix used by parse/context commands.
- If the expected `kodus-graph` version is already present, the worker skips unnecessary installs and proceeds to graph parsing.

### Acceptance Criteria

- The production worker image installs `@kodus/kodus-graph` for the runtime user or otherwise makes it available in that user's PATH.
- The Dockerfile pins the installed graph CLI version to the version expected by `KodusGraphCli`.
- `KodusGraphCli.install()` uses `BUN_PATH_PREFIX` when checking for an existing binary.
- Unit coverage verifies that a preinstalled `/home/node/.bun/bin/kodus-graph` with the expected version is detected and does not trigger another install.
- Existing parse/context commands continue to use `BUN_PATH_PREFIX`.

### Technical Details

Relevant files:

- `docker/Dockerfile`
- `libs/code-review/infrastructure/adapters/services/graph/kodus-graph-cli.ts`
- `test/unit/code-review/infrastructure/adapters/services/graph/graph-indexer.service.spec.ts`

### Evidence

On our self-hosted deployment, the published worker image needed a local runtime hotfix to make `kodus-graph` visible under `/home/node/.bun/bin` and to make the install check use the Bun global bin path.

The proposed fix moves that behavior into the upstream image/code path.

tulnovdl:
## Summary

Fixes #1343

The self-hosted worker image installed `@kodus/kodus-graph` before dropping privileges to `USER 1000`, but the worker process runs with `HOME=/home/node` and did not have `kodus-graph` available in its runtime PATH.

This PR makes the self-hosted worker image expose `kodus-graph` to the runtime user and aligns `KodusGraphCli.install()` with the existing parse/context commands by checking for the binary through `BUN_PATH_PREFIX`.

## What changed

- Add `KODUS_GRAPH_VERSION=0.2.19` to the production Dockerfile runtime stage.
- Add `/home/node/.bun/bin` to the runtime `PATH`.
- Install `@kodus/kodus-graph@${KODUS_GRAPH_VERSION}` after `USER 1000`, so Bun installs it under the runtime user's home.
- Update `KodusGraphCli.install()` to run the existing-binary check with `BUN_PATH_PREFIX`.
- Add regression coverage for detecting a preinstalled `/home/node/.bun/bin/kodus-graph` and skipping a redundant install.

## Changelog routing

`fix:` routes this change to the public **Bug fixes** changelog section.

## Test plan

- [x] `corepack pnpm exec prettier --write libs/code-review/infrastructure/adapters/services/graph/kodus-graph-cli.ts test/unit/code-review/infrastructure/adapters/services/graph/graph-indexer.service.spec.ts`
- [x] `corepack pnpm exec jest test/unit/code-review/infrastructure/adapters/services/graph/graph-indexer.service.spec.ts --runInBand --config jest.config.ts`
- [x] `git diff --check`

## Notes for reviewers

This is a self-hosted AST graph reliability fix. The Dockerfile already attempted to install `kodus-graph`, but the published worker image did not expose it to the non-root runtime user. The code path also had a small inconsistency: parse/context commands used `BUN_PATH_PREFIX`, while the install check did not.

The fix keeps the graph CLI version pinned to the same version expected by `KodusGraphCli`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -201,6 +201,10 @@ USER 1000
 # Install kodus-graph for the runtime user so self-hosted worker images can
 # find it through $HOME/.bun/bin after dropping root privileges.
 RUN bun install -g @kodus/kodus-graph@${KODUS_GRAPH_VERSION}
+USER root
+RUN printf '%s\n' '#!/bin/sh' 'exec /home/node/.bun/bin/kodus-graph "$@"' > /usr/local/bin/kodus-graph \
+  && chmod 0755 /usr/local/bin/kodus-graph
+USER 1000
 
 ENTRYPOINT ["/usr/local/bin/prod-entrypoint"]
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -158,12 +158,14 @@ ENV NODE_ENV=production
 # Redeclare ARGs to make them available in this stage
 ARG API_CLOUD_MODE=false
 ARG RELEASE_VERSION=local
+ARG KODUS_GRAPH_VERSION=0.2.19
 
 # Set ENVs for runtime
 ENV API_CLOUD_MODE=${API_CLOUD_MODE}
 ENV API_DEVELOPMENT_MODE=false
 ENV RELEASE_VERSION=${RELEASE_VERSION}
 ENV SENTRY_RELEASE=${RELEASE_VERSION}
+ENV PATH="/home/node/.bun/bin:${PATH}"
 
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt,sharing=locked \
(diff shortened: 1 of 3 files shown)


tulnovdl:
@kody start-review

Issue #1351, opened 2026-06-17
Inconsistent user.id type in pullRequests collection — saved as both string and number


sartorijr92:
Description

   In the MongoDB pullRequests collection, the nested user.id field is being persisted with inconsistent types: some
   documents store it as a string, others as a number. The same issue likely affects reviewers.id and assignees.id,
   although those paths already partially normalize the value.

   This makes queries, aggregations, and analytics that match or group by user.id unreliable, because MongoDB treats 123
   (number) and "123" (string) as different values.

   Root cause

   The user property is declared in the Mongoose schema as an untyped Object:

   ```ts
     // libs/platformData/infrastructure/adapters/repositories/schemas/pullRequests.model.ts
     @Prop({ type: Object, required: false })
     public user: {
         id: string;
         username: string;
     };
   ```

   Because the schema only enforces “this is an object”, Mongoose does not coerce user.id to a string on save.

   At the same time, the code that builds the user object before saving does not consistently convert id to string.
   Different Git providers return id in different native types:

   • GitHub → user.id is a number.
   • GitLab → user.id is a number.
   • Azure Repos → id is usually a string/UUID.
   • Bitbucket → UUID is already a string.
   • Forgejo → already converted to string.

   Affected code paths

   1. Main webhook save flow (extractUser / extractUsers)
       • File: libs/platformData/infrastructure/adapters/services/pullRequests.service.ts
       • id is assigned from data?.id, data?.uuid, or completeUser.id without String().
   2. Webhook payload mappers
       • libs/common/utils/webhooks/github.ts
       • libs/common/utils/webhooks/gitlab.ts
       • libs/common/utils/webhooks/forgejo.ts
       • These pass user directly from the provider payload into the PR object.
   3. Historical backfill
       • File: libs/platformData/application/use-cases/pullRequests/backfill-historical-prs.use-case.ts
       • user.id is assigned as pr.user?.id || '' without String().
       • Note: reviewers and assignees in the same function already use String(...), which shows the inconsistency.
   4. Mongoose schema
       • File: libs/platformData/infrastructure/adapters/repositories/schemas/pullRequests.model.ts
       • user is declared as Object, so there is no automatic type coercion.

   Expected behavior

   user.id, reviewers[].id, and assignees[].id should always be persisted as strings, regardless of the provider or code
   path.

   Suggested fix

   1. Normalize IDs in PullRequestsService.extractUser / extractUsers before returning:
     ```ts
[...]

PR #1423, opened 2026-07-01
fix: normalize user.id, reviewers[].id and assignees[].id as strings …

Nalu-devs:
#1351


libs/platformData/infrastructure/adapters/services/pullRequests.service.ts
```diff
@@ -1374,7 +1374,7 @@ export class PullRequestsService implements IPullRequestsService {
             }
 
             return {
-                id: data?.id || data?.uuid || '',
+                id: String(data?.id || data?.uuid || ''),
```
Wellington01, line 1377: @kody Why use String() instead of toString()?
If the ternary operator returns '', will String() work?

PR #1469, opened 2026-07-06
feat(sandbox): harden local sandbox readFile/writeFile path validation

Fanduzi:
## Scope

Harden local sandbox `readFile` and `writeFile` path validation to match the existing protections already applied to `remoteCommands` (grep, read, listDir, exec).

**Changes:**
- Reject absolute paths and `..` traversal in `readFile`/`writeFile`
- Reject symlink escapes for reads (via `resolveSafePath`)
- Reject writes through symlinked parent directories or symlinked target files (via new `resolveSafeWritePath`)
- Extract testable `buildSandboxFileAccess` helper matching the existing `buildRemoteCommands` pattern
- Add 10 focused unit+integration tests for normal access and escape attempts

**Files changed:**
- `libs/sandbox/infrastructure/providers/local-sandbox.service.ts` — added `resolveSafeWritePath`, `buildSandboxFileAccess`; replaced vulnerable inline closures
- `libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts` — added `describe('LocalSandboxService sandbox file access')` with 10 tests

## Non-goals

- Does NOT change sandbox lease cleanup behavior (→ local cleanup PR)
- Does NOT change E2B sandbox lifecycle, kill, or retry behavior
- Does NOT change reaper batching or concurrency (→ reaper concurrency PR)
- Does NOT introduce new lease states or Mongo state-machine changes

## Validation

```
✅ pnpm test -- --runTestsByPath libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts --no-coverage --runInBand (16/16 pass)
✅ pnpm exec prettier --check on changed files
✅ pnpm exec eslint on changed files
✅ pnpm run typecheck:libs-gate
```

## Pre-review Checklist

- [x] The diff does not touch lease repositories, reaper services, or invalidation flows
- [x] No cleanup retry state is introduced in this PR
- [x] Every new test asserts file-access behavior only
- [x] A reviewer asking about `/tmp/kodus-sandbox-*` cleanup should be redirected to the local cleanup PR

## Follow-ups

- Self-hosted `/tmp/kodus-sandbox-*` directory cleanup → local sandbox cleanup PR
- Reaper batching and E2B retry behavior → reaper concurrency PR

### DIFF
--- a/libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts
+++ b/libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts
@@ -190,4 +190,29 @@ describe('LocalSandboxService sandbox file access', () => {
             'ok',
         );
     });
+
+    it('rejects write when parent directory realpath escapes repo after mkdir (TOCTOU post-mkdir check)', async () => {
+        const fsPromises = require('fs/promises');
+        const originalRealpath = fsPromises.realpath;
+        // Track the nested subdir path — the post-mkdir finalCheck resolves
+        // the immediate parent of the target file.
+        const nestedDir = path.join(dir, 'escape', 'nested');
+
+        jest.spyOn(fsPromises, 'realpath').mockImplementation(
+            async (p: string) => {
+                if (p === nestedDir) {
+                    return '/tmp/outside-escape';
+                }
+                return originalRealpath(p);
+            },
+        );
+
+        try {
+            await expect(
+                sandbox.writeFile('escape/nested/file.txt', 'pwned'),
+            ).rejects.toThrow(/Path escapes repo boundary after mkdir/);
+        } finally {
+            jest.restoreAllMocks();
+        }
+    });
 });
--- a/libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts
+++ b/libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts
@@ -108,16 +108,28 @@ describe('LocalSandboxService sandbox file access', () => {
         );
     });
 
-    it('rejects absolute read paths', async () => {
+    it('rejects absolute read paths outside the repo', async () => {
         await expect(sandbox.readFile('/etc/passwd')).rejects.toThrow(
             /Absolute paths are not allowed/,
         );
     });
 
-    it('rejects absolute write paths', async () => {
+    it('rejects absolute write paths outside the repo', async () => {
         await expect(sandbox.writeFile('/etc/passwd', 'x')).rejects.toThrow(
             /Absolute paths are not allowed/,
         );
+    });
+
+    it('accepts absolute paths that resolve inside the repo (read)', async () => {
+        const absPath = path.join(dir, 'a.txt');
+        fs.writeFileSync(absPath, 'hello');
+        await expect(sandbox.readFile(absPath)).resolves.toBe('hello');
+    });
+
+    it('accepts absolute paths that resolve inside the repo (write)', async () => {
+        const absPath = path.join(dir, 'sub', 'abs.txt');
+        await sandbox.writeFile(absPath, 'content');
+        expect(fs.readFileSync(absPath, 'utf-8')).toBe('content');
     });
 
     it('rejects .. traversal reads', async () => {
--- a/libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts
+++ b/libs/sandbox/infrastructure/providers/local-sandbox.service.spec.ts
@@ -1,3 +1,7 @@
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+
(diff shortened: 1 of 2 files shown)


Fanduzi:
@malinosqui @sartorijr92

Here's where this PR stands and what I need your call on.

The local sandbox `readFile`/`writeFile` hardening covers: absolute paths only under repoDir, `..` traversal check before repoDir-prefix acceptance, symlink rejection on reads and writes, `O_NOFOLLOW` + `O_TRUNC` on the write target, and a post-mkdir `realpath` re-check.

Two things remain. First, a correctness fix I'll do now: the prefix checks use hardcoded `/` which breaks on Windows. I'll switch to `path.relative`.

Second, the write path still has a TOCTOU window between post-mkdir validation and `open()`. A concurrent actor on the same host could swap a validated parent for a symlink in that gap. Closing it properly needs `openat`-style fd operations — keep a parent dir fd open with `O_DIRECTORY | O_NOFOLLOW`, write relative to it. Node.js doesn't expose `openat` natively, so this means a native binding or accepting the current post-mkdir re-check as defense-in-depth.

Rather than keep layering patches on the write path, I'd rather get your call on the merge bar. If practical hardening with a documented residual is fine — I'll fix `path.relative`, document the window, and open a follow-up for fd-chain. If you want the full guarantee before merge, I'll pause this PR and scope that as a separate architecture change.

The read path is clean. The write path is where the question lives.

Wellington01:
Hi @Fanduzi — thanks for this, and for the thorough test coverage.

On the `resolveSafeWritePath` TOCTOU parent-swap finding (kody-ai, Security): it's technically valid — `O_NOFOLLOW` only guards the final path component, so a parent directory swapped for a symlink in the window between the post-`mkdir` `realpath` check and `open()` can still be traversed. We're accepting this as a documented, narrow residual rather than a blocker:

1. **Threat model.** Winning the race requires concurrent filesystem operations (symlink / `rm` / `mv`) inside the sandbox, timed to the write. The sandbox already exposes `remoteCommands.exec` (arbitrary shell), so an actor able to win this race can already write outside the repo directly — this hardening is not the last line of defense against an exec-capable agent. Its real value is containing **accidental** escapes and **path-string-only** attackers, which the current mitigations (double `realpath` around `mkdir` + `O_NOFOLLOW` on the final component + `..`-first rejection) cover well.

2. **The "clean" fix isn't a trivial Node change.** True `openat` semantics (open the validated parent with `O_DIRECTORY | O_NOFOLLOW`, then write relative to that fd) aren't available through Node's `fs` API — there's no `dirfd` variant of `fs.open`. Closing the window properly would require the Linux-only `/proc/self/fd/<fd>/<name>` trick or a native addon; the snippet suggested in the review (`open(parentFd, relativeFile, …)`) does not compile against Node's `fs`.

**Decision:** merging as-is. We'll revisit with a `/proc/self/fd`-based `openat` only if the sandbox ever runs untrusted code **without** an exec surface exposed. Tracking as a follow-up, not a blocker.

Minor / non-blocking: the `startsWith(repoReal + '/')` prefix checks are POSIX-only, but the local sandbox is Linux/`tmpdir`-based so there is no real Windows path here — no change required unless you'd like to switch to `path.relative` for robustness.

Fanduzi:
Thanks for the clear direction. I will leave this PR as-is and keep the openat/fd-chain work as a follow-up only if the sandbox threat model changes.

Wellington01:
**Decision: merging as-is — the residual TOCTOU is accepted and tracked, not a blocker.**

Recapping the rationale from the review discussion so it's on the record:

- The remaining parent-dir-swap race requires concurrent filesystem ops inside the sandbox timed to the write. The sandbox already exposes `remoteCommands.exec` (arbitrary shell), so an actor able to win this race can already write outside the repo directly — this hardening is defense-in-depth against **accidental** escapes and **path-string-only** attackers, not the last line of defense against an exec-capable agent.
- A proper fix needs true `openat` semantics (`O_DIRECTORY | O_NOFOLLOW` parent fd + write relative to it), which **Node's `fs` API doesn't expose** — the suggested `open(parentFd, relativeFile, …)` snippet doesn't compile. Closing it needs the Linux-only `/proc/self/fd` trick or a native addon.

Tracked as a follow-up: https://github.com/kodustech/kodus-ai/issues/1532 (revisit if the sandbox ever runs untrusted code **without** an exec surface, at which point the residual is no longer acceptable).

I'm dismissing the standing Kody `CHANGES_REQUESTED` on this basis so the merge gate reflects the decision. Thanks @Fanduzi for the thorough hardening + test coverage.

Wellington01 (approved):
Approving on the documented-residual basis (see decision comment + follow-up #1532). Hardening + tests look good; the remaining parent-dir-swap TOCTOU is accepted defense-in-depth, tracked separately.

PR #1471, opened 2026-07-06
feat(sandbox): add self-hosted local sandbox cleanup

Fanduzi:
## Scope

Clean up `/tmp/kodus-sandbox-*` directories for self-hosted local sandboxes on release, invalidation, creator failure, and reaper paths, while guarding every delete by expected `sandboxId` to prevent deleting fresh leases.

**Changes:**
- Add path-hardened local cleanup helper (`isLocalSandboxPath`, `deleteLocalSandbox`)
- Add cleanup status fields to lease model (`cleanupStatus`, `cleanupAttempts`, `cleanupRetryAt`, `cleanupError`)
- Add atomic claim/complete/fail repository methods guarded by `prKey + sandboxId`
- Clean local sandboxes on `release()` when `leaseCount` reaches zero
- Clean inactive local sandboxes immediately on `invalidate()`
- Clean local sandboxes on creator failure with retry marker
- Add local cleanup to reaper for expired and retryable leases
- Guard every delete by expected `sandboxId` to prevent deleting fresh leases
- Treat missing directories as success

**Files changed:**
- `libs/sandbox/infrastructure/repositories/schemas/sandbox-lease.model.ts` — added cleanup fields
- `libs/sandbox/infrastructure/services/local-sandbox-cleanup.service.ts` — NEW: path-hardened cleanup helper
- `libs/sandbox/infrastructure/services/local-sandbox-cleanup.spec.ts` — NEW: helper tests
- `libs/sandbox/infrastructure/repositories/sandbox-lease.repository.ts` — added atomic cleanup methods
- `libs/sandbox/infrastructure/services/sandbox-lease-manager.service.ts` — wired cleanup into release/invalidate/creator-failure
- `libs/sandbox/infrastructure/services/sandbox-lease-reaper.service.ts` — added local cleanup to reaper
- `libs/sandbox/infrastructure/services/sandbox-lease-manager.spec.ts` — added cleanup tests

## Non-goals

- Does NOT change local sandbox file access security (→ Goal 1 PR #1469)
- Does NOT change E2B kill/retry semantics
- Does NOT add reaper concurrency limiting (→ Goal 3 reaper concurrency PR)
- Does NOT perform broad lease state-machine refactoring
- Does NOT change file path traversal hardening (→ Goal 1 PR #1469)

## Validation

```
✅ pnpm test --runTestsByPath on 5 spec files (48/48 pass)
✅ pnpm exec prettier --check on changed files
✅ pnpm exec eslint on changed files
✅ pnpm run typecheck:libs-gate
✅ git diff --check
```

## Key Invariant

Every destructive filesystem operation validates both `os.tmpdir()` ancestry and the `kodus-sandbox-` basename prefix. Every lease delete or cleanup-complete operation is guarded by expected `sandboxId` via atomic `findOneAndUpdate`. A concurrent acquire creating a fresh lease for the same `prKey` will never be deleted by stale cleanup.

## Pre-review Checklist

- [x] Every destructive filesystem operation validates both `os.tmpdir()` ancestry and the `kodus-sandbox-` basename prefix
- [x] Every lease delete or cleanup-complete operation is guarded by expected `sandboxId`
- [x] Local cleanup treats missing directories as success
- [x] E2B kill behavior is unchanged in this PR
- [x] Reaper concurrency is unchanged in this PR
- [x] File path traversal hardening is unchanged in this PR

## Follow-ups

- Local sandbox path traversal hardening → Goal 1 PR #1469
- Reaper concurrency limiting and E2B already-gone handling → Goal 3 reaper concurrency PR

### DIFF
--- a/libs/sandbox/infrastructure/repositories/sandbox-lease.repository.ts
+++ b/libs/sandbox/infrastructure/repositories/sandbox-lease.repository.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { InjectModel } from '@nestjs/mongoose';
 import { Model } from 'mongoose';
 
+import { SANDBOX_LEASE_CLEANUP_STATUS } from './schemas/sandbox-lease.model';
 import { SandboxLeaseModel } from './schemas/sandbox-lease.model';
 
 /**
@@ -218,5 +219,83 @@ export class SandboxLeaseRepository {
             })
             .select('_id sandboxId killAt')
             .lean();
+    }
+
+    /**
+     * Atomically claim a local cleanup. Returns the doc only when prKey and
+     * sandboxId match and cleanup is not already in_progress or completed.
+     */
+    async claimCleanup(
+        prKey: string,
+        expectedSandboxId: string,
+    ): Promise<SandboxLeaseModel | null> {
+        return this.leaseModel
+            .findOneAndUpdate(
+                {
+                    _id: prKey,
+                    sandboxId: expectedSandboxId,
+                    cleanupStatus: {
+                        $nin: [
+                            SANDBOX_LEASE_CLEANUP_STATUS.IN_PROGRESS,
+                            SANDBOX_LEASE_CLEANUP_STATUS.COMPLETED,
+                        ],
+                    },
+                },
+                {
+                    $set: {
+                        cleanupStatus: SANDBOX_LEASE_CLEANUP_STATUS.IN_PROGRESS,
+                        cleanupRetryAt: null,
+                    },
+                    $inc: { cleanupAttempts: 1 },
+                },
+                { new: true },
+            )
+            .exec();
+    }
+
+    /**
+     * Complete a local cleanup by deleting the lease doc.
+     * Only deletes when prKey, sandboxId, and cleanupStatus=IN_PROGRESS match.
+     */
+    async completeCleanup(
+        prKey: string,
+        expectedSandboxId: string,
+    ): Promise<boolean> {
+        const result = await this.leaseModel
+            .deleteOne({
+                _id: prKey,
+                sandboxId: expectedSandboxId,
+                cleanupStatus: SANDBOX_LEASE_CLEANUP_STATUS.IN_PROGRESS,
+            })
+            .exec();
+        return result.deletedCount === 1;
+    }
+
+    /**
+     * Mark a local cleanup as failed. Sets a retry marker.
+     * Only updates when prKey, sandboxId, and cleanupStatus=IN_PROGRESS match.
+     */
+    async failCleanup(
+        prKey: string,
+        expectedSandboxId: string,
+        error: string,
+    ): Promise<boolean> {
+        const result = await this.leaseModel
+            .updateOne(
+                {
+                    _id: prKey,
+                    sandboxId: expectedSandboxId,
+                    cleanupStatus: SANDBOX_LEASE_CLEANUP_STATUS.IN_PROGRESS,
+                },
+                {
+                    $set: {
+                        cleanupStatus: SANDBOX_LEASE_CLEANUP_STATUS.FAILED,
+                        cleanupRetryAt: new Date(Date.now() + 60_000),
+                        cleanupError: error.slice(0, 500),
+                    },
+                },
+            )
+            .exec();
+        return result.modifiedCount === 1;
     }
 }
--- a/libs/sandbox/infrastructure/repositories/sandbox-lease.repository.ts
+++ b/libs/sandbox/infrastructure/repositories/sandbox-lease.repository.ts
@@ -128,12 +128,17 @@ export class SandboxLeaseRepository {
     }
 
     /**
-     * Mark a CREATING lease as INVALIDATED (mid-create race handling).
-     * The create path will check for this state after completing and kill the sandbox.
+     * Mark a CREATING or READY lease as INVALIDATED.
+     *
+     * CREATING: mid-create race handling — the create path detects this
+     * after completing and kills the sandbox.
+     * READY: called by invalidate() when a local READY lease has active
+     * consumers (leaseCount > 0). The state change prevents new acquires
+     * from attaching to the stale sandbox.
      */
     async markInvalidated(prKey: string): Promise<void> {
         await this.leaseModel.updateOne(
-            { _id: prKey, state: 'CREATING' },
+            { _id: prKey, state: { $in: ['CREATING', 'READY'] } },
             { $set: { state: 'INVALIDATED' } },
         );
     }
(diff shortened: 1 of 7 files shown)


Fanduzi:
@malinosqui @sartorijr92

The core cleanup functionality is in: persist temp dir as sandboxId, cleanup on release/invalidation/creator failure, guarded delete by expected sandboxId, stale retry, reaper path for local leases. Tests cover the happy paths and the guarded-delete invariant.

The problem isn't individual bugs — it's that I didn't fully define the state machine before implementing. `leaseCount` and `cleanupStatus` interact in ways that create deadlocks. Specifically: `acquire()` increments `leaseCount` then waits on `cleanupStatus=IN_PROGRESS`, but `claimCleanup` needs `leaseCount <= 0` to proceed. Each retry inflates the count. Separately, when cleanup fails in the creator path, the doc is preserved but `leaseCount` isn't decremented, so the reaper skips it.

I can do a focused repair: define the invariant (who decrements in each failure path, when cleanup can proceed relative to leaseCount), fix the two deadlock paths, add tests to pin it. That stays within this PR's scope but it's a real design change, not a patch.

Alternatively, I close this PR, write a short design doc for the cleanup state machine, and reopen after we agree on invariants.

I'd prefer doing the repair here if you're comfortable with the scope. But if you'd rather see the design first, I'll do that. Either way, I'd rather get your direction than keep throwing patches at an undefined state machine.

Wellington01:
Hi @Fanduzi — the architecture here is solid: atomic `claimCleanup` guarded by `_id + sandboxId + status`, the `requireLeaseCountZero` guard against a racing acquire, and `resetStaleCleanup` for crashed-worker recovery. Most of the earlier review points are already addressed in the current head.

One genuine issue left before we merge — the **acquire-vs-invalidated-local** race (also flagged by codex):

When `invalidate()` flips a READY **local** sandbox with active consumers to `INVALIDATED` (the `leaseCount > 0` branch → `markInvalidated`), a new `acquire()` for the same `prKey` runs `upsertAcquire()` (which increments `leaseCount`) and only *then* throws in `handleJoinerPath()` on the `INVALIDATED` state — **before** any `leaseId` is tracked. That increment is never released, so `leaseCount` stays `> 0`, local cleanup is skipped, and the temp directory lingers until the TTL while new reviews keep failing on the invalidated doc.

Could you make acquires bypass/delete an `INVALIDATED` local lease (or decrement on the failed-acquire path) so the count can't leak? Once that's in, we're good to merge.

Heads-up on ordering: this PR and #1473 both modify `reaper.service.ts` (`Promise.allSettled` → local cleanup here vs. `mapWithConcurrency` there), so they will conflict — whichever lands second will need a rebase.

Fanduzi:
@Wellington01 Fixed. The acquire-vs-invalidated-local race is addressed in `acquire()` — after `upsertAcquire()` increments `leaseCount`, if the doc has `state === INVALIDATED`, the code now decrements the increment and throws. The doc is preserved (not deleted) so an in-flight creator can still observe the INVALIDATED state. The caller gets a clear error and can retry with a fresh acquire if needed.

Two Codex findings came up on the latest push, but both are on pre-existing code paths (not from this fix):
- `invalidate()` race when `claimCleanup` loses to a concurrent acquire (line 433)
- cleanup IN_PROGRESS retry path not decrementing leaseCount (line 184)

These are follow-ups, not blockers for this PR.

Re: ordering — noted on the reaper.service.ts conflict with #1473. Happy to rebase whichever lands second.

Wellington01:
Hi @Fanduzi — #1473 just merged, which means this PR is now **CONFLICTING** on `sandbox-lease-reaper.service.ts` (the collision I flagged earlier). Here's exactly what changed and how to resolve it so we can get this in.

### What changed on `main`

#1473 replaced the unbounded `Promise.allSettled(expired.map(...))` with a bounded `mapWithConcurrency(...)` helper in **both** cron loops:

- `reapExpiredLeases()` → now `await mapWithConcurrency(expired, CLEANUP_CONCURRENCY, async (lease) => { ... })`
- `killIdleSandboxes()` → now `await mapWithConcurrency(ready, CLEANUP_CONCURRENCY, async (lease) => { ... })`

This PR was written against the old `Promise.allSettled` version and inserts the local-cleanup branch (`if (isLocalSandboxPath(lease.sandboxId)) await this.cleanupLocalLease(lease)`) into those loops, so git can't auto-merge.

### How to resolve

1. **Rebase `fix/local-sandbox-cleanup` onto latest `main`.**
2. In the conflict, **keep `main`'s `mapWithConcurrency(...)` structure** (don't restore `Promise.allSettled`) and re-insert your local-cleanup branch **inside the existing `mapWithConcurrency` callbacks**, in both crons. Concretely, at the top of each `async (lease) => { ... }` body:
   ```ts
   async (lease) => {
       // Local sandboxes have no remote E2B slot to free — clean the dir instead.
       if (isLocalSandboxPath(lease.sandboxId)) {
           await this.cleanupLocalLease(lease);
           return;
       }
       // ... existing E2B kill/delete logic from main ...
   }
   ```
3. Your new `cleanupLocalLease()` method and the `local-sandbox-cleanup.service` import are pure additions — they **don't conflict**, keep them as-is.

Net effect: the local-cleanup path now inherits the same bounded concurrency as the E2B path — which is actually the outcome we want (one code path, one concurrency budget).

### After the rebase

- The `leaseCount`-leak fix on the acquire-vs-INVALIDATED race (commit `3a9a3f4c`) already addresses **both** my ask and Kody's follow-up (the recursive-retry stack concern) by switching to decrement-and-throw. Once the rebase is up and CI is green, re-request review and I'll clear the standing `CHANGES_REQUESTED`.

### One non-blocking follow-up (your call)

With decrement-and-throw, the `INVALIDATED` doc isn't deleted on the failing acquire — it lingers until its **original** `expiresAt` elapses and the reaper sweeps it (`markInvalidated` only flips the state, it doesn't shorten the TTL). So the leak is fixed, but the "new acquires for that prKey keep throwing until the doc expires" window can stay open a while. If you want to close it, have `invalidate()`/`markInvalidated()` also set a short `expiresAt` so the reaper reclaims the doc quickly. Fine to do as a follow-up PR — not a blocker for this one.

Ping me once it's rebased and I'll review.

Fanduzi:
Hi @Wellington01 — before pushing another change, I reviewed the earlier discussion and want to confirm the boundary for this PR.
Your previous guidance was that the core local cleanup architecture was solid, the acquire-vs-INVALIDATED count leak was the merge blocker, and shortening INVALIDATED TTL was a non-blocking follow-up. The new review findings now touch several related lease/cleanup races:
- an acquire that sees cleanup IN_PROGRESS can leave a phantom leaseCount increment;
- expired local leases can be skipped when a crashed worker left leaseCount > 0;
- invalidate() can lose its cleanup claim to a concurrent acquire and leave a READY local lease usable;
- a FAILED cleanup marker can be reused as a READY sandbox;
- completeCleanup can race with a reacquire after the cleanup claim.
I can make one focused pass on these cleanup-state invariants, keeping the changes inside SandboxLeaseManager, SandboxLeaseRepository, the local cleanup helper, and the reaper. I would leave the direct-import/DI refactor and the INVALIDATED TTL policy as follow-up work.
I have not pushed the local draft yet. Does this scope belong in #1471, or should some of these be separate follow-ups?

Wellington01:
Hey @Fanduzi — good call checking scope before pushing more. I went through the current head (`328344c57`) against main, and honestly it's not five things. Three of them are already handled in your branch, so let's not go in circles on those:

- the acquire-vs-INVALIDATED leak (my original blocker) is done — decrement-and-throw in `acquire()` at `sandbox-lease-manager.service.ts:194`, doc kept around
- `markInvalidated` now covers READY too (`repository.ts:141`, `$in: ['CREATING','READY']`) — that was the Codex P2
- `claimCleanup` has the lease-count guard and release/invalidate both pass `true` — that was the Kody/Codex P1

And two of the five you listed (#1 and #5) are basically already covered by the IN_PROGRESS poll you added in `acquire()` (`:172`) — a racing acquire just parks in the poll loop and never touches the doc that's being deleted, and its `$inc` goes away with the `completeCleanup` delete. I'd leave those alone, piling more guards on there just makes it harder to read for no real gain.

So what's actually left is three, and they're all the same thing you called out on the 7th — `leaseCount` / `cleanupStatus` / `state` are an implicit state machine nobody wrote down. Each one is just a different corner of it:

1. reaper skips crashed-worker leases — `cleanupLocalLease` bails when `current.leaseCount > 0` (`reaper.service.ts:229`), but `findExpired` is explicitly supposed to return those (`repository.ts:155`, "crashed-worker leases stay at leaseCount > 0 forever"). So expired + dead worker = the dir just sits in /tmp forever. This is the real one.
2. `invalidate()` can lose the claim — the `leaseCount <= 0` branch (`:429`) reads the count and then `claimCleanup(true)` loses the race to an acquire that re-increments. Claim fails, no fallback, and the READY local lease ends up neither cleaned nor invalidated.
3. `acquire()` never rejects `FAILED` — `failCleanup` leaves `state = READY` + `sandboxId` and just flips `cleanupStatus` to `FAILED` (`:296`), but `acquire()` only checks `IN_PROGRESS` and `INVALIDATED`, so the next acquire happily connects to a half-cleaned dir.

My take: keep it in this PR. It's three races, all in the four files you're already touching, same root cause — splitting them into follow-ups just ships known races. But let's kill the patch→review→patch loop the way you already suggested: write the invariant down first. Four or five lines at the top of `SandboxLeaseManager` — who decrements `leaseCount` on each failure path, when cleanup is allowed to run vs `leaseCount`, and that `acquire()` has to reject both `FAILED` and `INVALIDATED`. Once that's there the three fixes basically fall out of it and I can review against the spec instead of guessing case by case.

DI/import refactor and the INVALIDATED TTL thing — agreed, those are separate PRs, leave them out.

Do those three + the invariant, ping me for re-review and I'll drop the `CHANGES_REQUESTED`.

Fanduzi:
Hi @Wellington01 — pushed the focused state-machine pass in commit bc2960c33, following your scope:
- expired local leases now force-reclaim even when a crashed worker left leaseCount greater than zero; idle cleanup still requires zero leases;
- if invalidate() loses the guarded cleanup claim to a concurrent acquire, it re-reads the same READY local lease and marks it INVALIDATED;
- acquires now reject FAILED local cleanup leases and balance the acquire count instead of reconnecting to the failed directory;
- added the lease invariants above SandboxLeaseManager and focused regression tests.
Validation: 59 sandbox tests passed; Prettier and ESLint passed. I left the IN_PROGRESS polling, DI refactor, and INVALIDATED TTL policy unchanged. Please re-review this head when available.

PR #1491, opened 2026-07-07
fix(kody-rules): match comma-joined multi-glob rule paths at review time

### LINKED ISSUE #1483: Kody Rules with comma-joined multi-glob path never match any file at review time
## Summary
A Kody Rule whose `path` contains multiple comma-joined globs (e.g. `app/**/*.rb,lib/**/*.rb`) is silently never applied to any file during review. Single-glob rules work normally.

## Root cause
The IDE-rule importer explicitly comma-joins declared globs (prompt instruction "Comma-join multiple declared globs" in `libs/kodyRules/infrastructure/adapters/services/kodyRulesSync.service.ts:1850`), but the review-time matcher passes the whole string to picomatch as a single pattern:

- `isFilePathMatch` → `isFileMatchingGlob(normalizedFilename, [rulePath])` (`libs/ee/kodyRules/service/kody-rules-validation.service.ts:445`)
- picomatch treats the comma as a literal character, so the pattern matches zero files (verified: `picomatch('app/**/*.rb,lib/**/*.rb')('app/models/user.rb') === false`; the braced form `{a,b}` matches).

Parts of the codebase assume comma-joined globs ARE supported — `pathMatchesIdeRuleDir` splits on comma with the comment "KodyRules supports OR-joined globs" (`libs/common/utils/kody-rules/file-patterns.ts:154`). The review matchers never got that support.

## Impact
Reported by a self-hosted customer: 5 of 9 imported rule files (all multi-glob) produced zero findings across two controlled experiments with planted violations, while all 4 single-glob rules fired with 100% recall. Rules appear ACTIVE in the UI with no indication they can never match.

## Fix suggestion
Split `rulePath` on comma in `isFilePathMatch` and `isFolderPathMatch` before calling `isFileMatchingGlob` (which already accepts an array). Fixing on the read path repairs all already-persisted rules without a migration. `isFolderPathMatch` also needs the split before `getGlobBasePath`.

malinosqui:
## Problem

A Kody Rule whose `path` contains multiple comma-joined globs (e.g. `app/**/*.rb,lib/**/*.rb`) was silently never applied to any file during review. The IDE-rule importer explicitly comma-joins declared globs, but `isFilePathMatch`/`isFolderPathMatch` passed the whole string to picomatch as a single pattern — where a comma is a literal character — so the pattern matched zero files. Single-glob rules worked, which produced a confusing all-or-nothing failure pattern per rule file (reported by a self-hosted customer with controlled planted-violation experiments).

`pathMatchesIdeRuleDir` already splits on commas ("KodyRules supports OR-joined globs"); the review-time matchers never got that support.

## Fix

- New `splitRulePathGlobs`: splits the rule path on commas, ignoring commas inside brace alternations (`{app,lib}/**` stays intact), trimming whitespace.
- `isFilePathMatch` ORs the resulting globs via `isFileMatchingGlob` (already array-aware).
- `isFolderPathMatch` runs its existing base-path logic per glob (`isFolderGlobMatch`).

Fixing on the read path repairs all already-persisted multi-glob rules with no migration.

## Tests

6 new tests in `kody-rules-validation.service.spec.ts`: multi-glob file match (both sides), non-match outside all globs, whitespace tolerance, brace alternation preservation, single-glob regression, folder matching. Full `kodyRules|kody-rules` suite: 42 suites / 394 tests green.

Fixes #1483

### DIFF
--- a/libs/ee/kodyRules/service/kody-rules-validation.service.ts
+++ b/libs/ee/kodyRules/service/kody-rules-validation.service.ts
@@ -418,6 +418,41 @@ export class KodyRulesValidationService {
     }
 
     /**
+     * Splits a rule's `path` into its individual glob patterns.
+     *
+     * Kody Rules support OR-joined globs in a single `path` string — the
+     * IDE-rule importer comma-joins declared globs (e.g. a frontmatter
+     * `path: ["app/**", "lib/**"]` is persisted as "app/**,lib/**") and
+     * `pathMatchesIdeRuleDir` has always split on commas. The review-time
+     * matchers below, however, used to pass the whole string to picomatch
+     * as ONE pattern, where a comma is a literal character — so any
+     * multi-glob rule silently matched zero files.
+     *
+     * Commas inside braces are NOT separators: "{app,lib}/**" is a single
+     * valid picomatch alternation and must stay intact.
+     */
+    private splitRulePathGlobs(rulePath: string): string[] {
+        const globs: string[] = [];
+        let current = '';
+        let braceDepth = 0;
+
+        for (const char of rulePath) {
+            if (char === '{') braceDepth++;
+            else if (char === '}' && braceDepth > 0) braceDepth--;
+
+            if (char === ',' && braceDepth === 0) {
+                if (current.trim()) globs.push(current.trim());
+                current = '';
+                continue;
+            }
+            current += char;
+        }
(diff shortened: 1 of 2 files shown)


malinosqui:
Superseded by #1494 (consolidated hotfix off tag 2.1.27, includes the review fix for character-class commas).

PR #1493, opened 2026-07-07
fix(kody-rules): detect root AGENTS.md and discover nested rule files

### LINKED ISSUE #1484: Root AGENTS.md is not detected as a rule file (only .agents.md/.agent.md dotfiles match)
## Summary
`RULE_FILE_PATTERNS` (`libs/common/utils/kody-rules/file-patterns.ts:5`) includes `.agents.md` and `.agent.md` (hidden dotfiles) but not the de-facto standard `AGENTS.md` at the repo root, so repos using the standard convention are never synced.

Case-insensitivity doesn't help: the leading dot means `.agents.md` can never match `AGENTS.md`.

## Related inconsistency
Full-repo discovery matches case-insensitively (`isFileMatchingGlobCaseInsensitive`, e.g. `github.service.ts:5570`), but the PR-driven sync path matches case-sensitively (`isFileMatchingGlob` in `kodyRulesSync.service.ts:390`). A lowercase `claude.md` is picked up by full sync but ignored by PR-driven sync. The two paths should agree.

## Fix suggestion
Add `AGENTS.md` (and consider `AGENT.md`) to `RULE_FILE_PATTERNS`, and unify the two sync paths on case-insensitive matching. Consider making the pattern list config-extensible so self-hosters don't need to fork the image to add a pattern.

### LINKED ISSUE #1485: Nested CLAUDE.md / rule files in subdirectories are not discovered by default
## Summary
Rule-file discovery matches `RULE_FILE_PATTERNS` against the full path, so root-level patterns like `CLAUDE.md` only match at the repo root. `sub/dir/CLAUDE.md` is only discovered if that directory is explicitly configured in the code review settings (`getDirectoryPatterns` prefixes patterns per configured directory, `kodyRulesSync.service.ts`).

Repos that split guidance into per-directory CLAUDE.md/AGENTS.md files (a common monorepo convention) silently lose everything outside the root.

Note the `**/`-prefixed variants already exist in `isIdeRuleSource` (`file-patterns.ts:65`) but only for recognizing existing sourcePaths, not for discovery.

## Fix suggestion
Either discover `**/CLAUDE.md` (and similar root-level patterns) by default, or document clearly that nested rule files require configuring each directory. `extractRepoSubdirFromIdeSource` already scopes nested sources to their subdirectory, so the scoping half is done.

malinosqui:
## Problem

Two customer-reported gaps in IDE rule-file detection (#1484, #1485):

1. The de-facto standard `AGENTS.md` (uppercase, repo root, no leading dot) never matched — `RULE_FILE_PATTERNS` only listed the `.agents.md`/`.agent.md` dotfiles — so repos following the agents.md convention were never synced.
2. Root-level patterns (`CLAUDE.md`, `AGENTS.md`, etc.) were matched against the full path, so per-directory guidance files (the common monorepo convention) were only discovered if each directory was explicitly configured in code review settings.

Bonus inconsistency: PR-driven sync matched rule files case-sensitively while full-repo listings matched case-insensitively, so a lowercase `claude.md` synced on full sync but was ignored on PR merge.

## Fix

- Add `AGENTS.md` to `RULE_FILE_PATTERNS`.
- New `RULE_FILE_DISCOVERY_PATTERNS` (every base pattern + its `**/` variant) used by all four discovery/filter sites in `KodyRulesSyncService`. This is the same list `isIdeRuleSource` already recognised, so anything the sync imports is guaranteed to be recognised as IDE-synced afterwards (toggle-off purge, orphan chip, etc.).
- Nested sources keep being scoped to their subdirectory by the existing `extractRepoSubdirFromIdeSource` — a `services/foo/CLAUDE.md` rule applies to `services/foo/**`, not repo-wide.
- Both sync paths now match with `isFileMatchingGlobCaseInsensitive`, aligning with the provider-side repo listings.

## Behavior change note

Existing installs will start discovering nested guidance files on the next sync. Rules stay scoped per subdirectory, and the onboarding fast-sync caps (maxFiles/maxBytes) still apply.

## Tests

9 new tests in `kody-rules-file-patterns.spec.ts` (root AGENTS.md, nested discovery, case-insensitivity, discovery-list invariant, subdir scoping). Full `file-patterns|kody-rules|kodyRules` sweep: 42 suites / 396 tests green.

Fixes #1484
Fixes #1485

### DIFF
--- a/libs/common/utils/kody-rules/file-patterns.ts
+++ b/libs/common/utils/kody-rules/file-patterns.ts
@@ -12,6 +12,7 @@ export const RULE_FILE_PATTERNS = [
     '.github/instructions/**/*.instructions.md',
 
     // Agentic
+    'AGENTS.md',
     '.agents.md',
     '.agent.md',
 
@@ -41,6 +42,23 @@ export const RULE_FILE_PATTERNS = [
 export type RuleFilePattern = (typeof RULE_FILE_PATTERNS)[number];
 
 /**
+ * Patterns used to DISCOVER rule files anywhere in a repository: every
+ * base pattern plus its `**\/`-prefixed variant, so root-level names like
+ * `CLAUDE.md` or `AGENTS.md` are also found in subdirectories (the
+ * common monorepo convention of per-directory guidance files).
+ *
+ * Discovery and recognition (`isIdeRuleSource`) intentionally share this
+ * list — a file the sync imports must always be recognised as IDE-synced
+ * afterwards. Nested sources are scoped to their subdirectory by
+ * `extractRepoSubdirFromIdeSource`, so a `services/foo/CLAUDE.md` rule
+ * applies to `services/foo/**`, not repo-wide.
+ */
+export const RULE_FILE_DISCOVERY_PATTERNS: ReadonlyArray<string> = [
+    ...RULE_FILE_PATTERNS,
+    ...RULE_FILE_PATTERNS.map((p) => `**/${p}`),
+];
+
+/**
  * Whether `sourcePath` came from an IDE rule file recognised by the auto-sync
  * importer (i.e. it matches one of `RULE_FILE_PATTERNS`, anywhere in the repo).
  *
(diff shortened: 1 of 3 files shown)


malinosqui:
Superseded by #1494 (consolidated hotfix off tag 2.1.27, includes the review fixes: case-insensitive isIdeRuleSource + AGENTS.md in the web infer-origin classifier).

PR #1495, opened 2026-07-07
feat(kody-rules): import .kody/rules template files verbatim (no LLM)

### LINKED ISSUE #1487: .kody/rules files should be imported verbatim (LLM conversion drops examples and rewrites content)
## Summary
All rule files, including `.kody/rules/**` (our own documented structured template), go through an LLM conversion step (gpt-4o-mini, `kodyRulesSync.service.ts:1820`) whose prompt merges each file into exactly one rule, keeps only 1-2 "minimal" examples, reformats markdown, and rewrites in English.

For free-form sources (CLAUDE.md, .cursorrules) that's reasonable. For `.kody/rules` the user already wrote the exact structure we ask for (YAML frontmatter with title/path/severity + Instructions + Bad/Good examples), and the conversion:
- drops the Bad/Good example sections,
- strips rule identifiers the user relies on (e.g. bold markers),
- reformats bullets non-deterministically.

The result is duplicate maintenance and rules that differ from the reviewed source of truth.

## Fix suggestion
Add a deterministic parser path for `.kody/rules/**`: parse frontmatter (title, path array, severity, scope), copy the body verbatim into the rule, and map examples structurally. Keep the LLM path only for free-form formats. Note: the frontmatter `path` is an array — see #1483 for how multi-glob paths must be stored/matched.

malinosqui:
## Problem

`.kody/rules/**` files follow our own documented structured template, but the importer ran them through the same LLM conversion used for free-form sources (CLAUDE.md, .cursorrules). The conversion merges everything into one condensed rule: Bad/Good examples were trimmed, wording rewritten, and rule identifiers (e.g. `**NAME1**`) stripped — reported by a self-hosted customer who authored 9 rule files exactly per the docs and found the imported content mangled.

## Fix

New deterministic parser `libs/common/utils/kody-rules/kody-rule-file-parser.ts`:

- Frontmatter: `title` (required), `path` (string or array → comma-joined, the storage form the review matchers consume), `severity_min` (documented) or `severity` (alias), `scope`, `uuid` passthrough, `enabled: false` → skip import.
- Body after the frontmatter becomes the rule text **verbatim**.
- Bad/Good examples extracted structurally (headings containing bad/incorrect/wrong vs good/correct; all fenced blocks under each).
- Returns `null` for anything that doesn't parse as the template → caller falls back to the LLM importer, so free-form `.kody` files keep working.

Wired into both import paths: `convertFileToKodyRules` (main/PR-driven sync) and the onboarding fast batch (templates are parsed locally, only free-form files go to the batch LLM). Declared template paths keep the `validateAndScopeIdeRulePath` guard so a template can't scope a rule against the rule sources themselves.

## Tests

19 new tests (`kody-rule-file-parser.spec.ts`) including the customer's exact template shape (multi-glob path array, `**CF1**` identifiers surviving, Bad/Good extraction) and fallback cases. Full kody sweep: 460 tests green.

Note: multi-glob comma-joined paths depend on the matcher fix in #1494.

Closes #1487

### DIFF
--- a/libs/common/utils/kody-rules/kody-rule-file-parser.ts
+++ b/libs/common/utils/kody-rules/kody-rule-file-parser.ts
@@ -1,0 +1,191 @@
+import * as yaml from 'js-yaml';
+
+/**
+ * Deterministic parser for `.kody/rules/**` markdown files (the structured
+ * template documented in "Repository Rules").
+ *
+ * These files are authored against OUR OWN template (YAML frontmatter +
+ * `## Instructions` + Bad/Good examples), so they must be imported
+ * verbatim — no LLM conversion. The LLM path (see
+ * `KodyRulesSyncService.convertFileToKodyRules`) remains the fallback for
+ * free-form sources (CLAUDE.md, .cursorrules, ...) and for `.kody` files
+ * that don't parse as the template.
+ *
+ * Returns `null` when the content doesn't look like the template
+ * (no/invalid frontmatter or no title), so callers can fall back to the
+ * LLM importer instead of failing the sync.
+ */
+
+export interface ParsedKodyRuleFile {
+    uuid?: string;
+    title: string;
+    /** Full markdown body after the frontmatter, verbatim. */
+    rule: string;
+    /** Comma-joined globs (the storage form the matchers consume). */
+    path: string;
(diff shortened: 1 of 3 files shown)


malinosqui:
Folded into the consolidated hotfix #1494 (single branch off 2.1.27) per release plan.

PR #1496, opened 2026-07-07
fix(kody-rules): stop resurrecting stale soft-deleted rules on re-import

malinosqui:
## Problem

Customer-reported (first report, item 5): an old rule persisted; toggling sync off auto-removed it, toggling on re-added it on the next run.

Root cause: toggle-off "delete" is a soft-delete (status → DELETED, record stays in the rules array), and `findRuleBySourcePath` matched with `Array.find` over ALL rules — soft-deleted included — returning the **first (oldest)** record for a sourcePath. When a stale deleted record coexisted with a newer live one, every re-import matched the stale uuid, and `createOrUpdate` merged the fresh content into it and flipped it back to ACTIVE — a resurrection of the oldest record, not an update of the current one.

## Fix

`findRuleBySourcePath` now prefers the **newest non-deleted** match. Falling back to a deleted record when it's the only match is kept intentionally: if the source file still exists in the repo, reviving that record (with content refreshed from the file) beats accumulating duplicates.

## Tests

4 new tests covering: newest-live wins over older-deleted, live wins even when older, sole-deleted fallback (revive), and no-match. Full kody sweep green.

### DIFF
--- a/libs/kodyRules/infrastructure/adapters/services/kodyRulesSync.service.ts
+++ b/libs/kodyRules/infrastructure/adapters/services/kodyRulesSync.service.ts
@@ -236,11 +236,34 @@ export class KodyRulesSyncService {
             const existing = await this.kodyRulesService.findByOrganizationId(
                 organizationAndTeamData.organizationId,
             );
-            const found = existing?.rules?.find(
-                (r) =>
-                    r?.repositoryId === repositoryId &&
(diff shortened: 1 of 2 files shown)


malinosqui:
Folded into the consolidated hotfix #1494 (single branch off 2.1.27) per release plan.

PR #1498, opened 2026-07-07
test(e2e): kody-rules-file-sync scenario (verbatim import + multi-glob enforcement)

malinosqui:
New E2E scenario modeling the self-hosted customer case behind #1483/#1487, so the whole chain is regression-guarded in the matrix:

1. Opens + **merges** a PR adding `.kody/rules/e2e-file-sync.md` — the documented template with `@kody-sync`, a `**E2ESYNC1**` identifier and a **multi-glob** `path` list.
2. Asserts the synced rule: imported **verbatim** (identifier survives), path stored comma-joined in declared order, status active.
3. Opens a violation PR planting the marker in a file matching the rule's **second** glob (the exact case the comma-join matcher bug silently killed) and asserts the review flags it (suggestion linked to the synced rule uuid, or review sample mentions the marker; one re-trigger for propagation races, same discipline as kody-rules-create-and-apply).

Details:
- Fixed rule-file path, overwritten each run — no fixture accumulation in the test repo; platform rule deleted in `finally`.
- github-only for now: repo-file sync fires on **merged** PRs and github is the only provider implementing `mergePR`.
- Registered in `sh-all`, `full`, `full-no-sso` matrices.
- Runs green only against builds containing the fixes from #1494 (that's the point).

### DIFF
--- a/tests/e2e/matrix/full-no-sso.yml
+++ b/tests/e2e/matrix/full-no-sso.yml
@@ -23,6 +23,10 @@ scenarios:
     # from the command. Covers all 4 providers.
     - command-review
     - kody-rules-create-and-apply
+    # Repo-file rule sync (customer case): .kody/rules template imported
+    # verbatim + multi-glob path enforced on a file matching the 2nd glob.
+    # github-only (needs mergePR).
+    - kody-rules-file-sync
     - license-attribution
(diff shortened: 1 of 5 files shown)


malinosqui:
Folded into #1494 — everything rides the single hotfix branch.

PR #1609, opened 2026-07-27
fix(gitlab): treat work_in_progress as draft signal on older GitLab (#1514)

### LINKED ISSUE #1514: GitLab Draft MRs can bypass runOnDraft=false when only work_in_progress is set
## Summary

Kodus currently determines whether a GitLab merge request is a draft by reading only the `draft` field. On at least GitLab self-managed 13.x, a Draft MR can be returned with `draft: null` and `work_in_progress: true`.

That makes Kodus map the MR as `isDraft=false`, so the `runOnDraft=false` guard is bypassed and an automatic review can run on a Draft MR.

## Confirmed example

A GitLab MR with a title prefixed by `Draft:` was returned by the GitLab API as:

```json
{
  "iid": 8423,
  "title": "Draft: ...",
  "state": "opened",
  "draft": null,
  "work_in_progress": true
}
```

With `runOnDraft=false`, this MR was still reviewed automatically.

## Affected versions

Confirmed:

- GitLab self-managed 13.x

Suspected:

- Older/self-managed GitLab versions around the WIP-to-Draft transition, where `work_in_progress` may still be the reliable draft indicator and `draft` may be missing or `null`.

Unknown:

- Exact lower/upper version boundary. The fix should be field-compatible rather than version-gated.

Relevant GitLab docs:

- Merge Request API still documents `work_in_progress` as deprecated and says to use `draft` instead: https://docs.gitlab.com/api/merge_requests/
- Webhook docs list `object_attributes.work_in_progress` as deprecated with `object_attributes.draft` as the recommended alternative: https://docs.gitlab.com/user/project/integrations/webhook_events/
- GitLab draft history notes WIP MRs were renamed to Draft in GitLab 13.2: https://docs.gitlab.com/user/project/merge_requests/drafts/

## Current code paths

The type definitions include both fields, but runtime code only uses `draft`:

- `libs/common/utils/webhooks/gitlab.ts`: webhook MR mapping uses `draft` only.
- `libs/platform/infrastructure/adapters/services/gitlab.service.ts`: API MR transform uses `mergeRequest?.draft ?? false`.
- `libs/platform/infrastructure/webhooks/gitlab/gitlabPullRequest.handler.ts`: Draft -> Ready detection watches only `changes.draft`.

Resulting chain:

```text
GitLab API returns draft=null, work_in_progress=true
  -> Kodus maps isDraft=false
  -> validate-config sees isDraft=false
  -> runOnDraft=false guard does not skip
  -> Draft MR is reviewed automatically
```

## Expected behavior

When GitLab returns `draft` as `null` or omits it, Kodus should fall back to `work_in_progress`:

```ts
isDraft = mergeRequest?.draft ?? mergeRequest?.work_in_progress ?? false;
```

## Suggested fix

- Update GitLab webhook mapping to use `draft ?? work_in_progress ?? false`.
- Update GitLab API MR transform to use `draft ?? work_in_progress ?? false`.
- Update Draft -> Ready detection to also handle `changes.work_in_progress: true -> false`.
- Update GitLab webhook types so `draft` and `work_in_progress` can be optional/null where GitLab may omit or null them.
- Add regression tests for `draft: null` + `work_in_progress: true`.

One edge case to consider: Draft -> Ready detection currently runs after the `changes.description` early return. If a GitLab webhook includes both description and draft-status changes, that early return may still suppress the ready transition.

vaibzde:
## Summary

Fixes #1514.

On GitLab self-managed 13.x, a **Draft** merge request can be returned by the API/webhook with `draft: null` **and** `work_in_progress: true`. Kodus decided draft state by reading only `draft`, so such MRs were mapped as `isDraft = false`. The `runOnDraft=false` guard was then bypassed and an automatic review ran on a Draft MR.

GitLab renamed "WIP" to "Draft" in 13.2. `work_in_progress` is deprecated but is still the reliable draft signal on those older instances, so the fix is **field-compatible** rather than version-gated.

## Root cause

Three sites keyed draft state off `draft` alone:

1. Webhook mapping — `libs/common/utils/webhooks/gitlab.ts` (`mapPullRequest`)
2. API transform — `libs/platform/infrastructure/adapters/services/gitlab.service.ts`
3. Draft→Ready transition detection — `libs/platform/infrastructure/webhooks/gitlab/gitlabPullRequest.handler.ts` (watched only `changes.draft`)

The `isDraft` set by (1)/(2) flows into the review gate at `libs/code-review/pipeline/stages/validate-config.stage.ts` (`if (isDraft && !runOnDraft) { skip }`), which is exactly the guard the issue reports as bypassed.

## Fix

Resolve draft state as `draft ?? work_in_progress ?? false`. This trusts an explicit `draft: false` when GitLab sends it (draft is canonical when present) and only falls back to the deprecated `work_in_progress` when `draft` is `null`/`undefined`.

- New shared helper `resolveGitlabIsDraft()` in `libs/common/utils/webhooks/gitlab.ts`, used by both the webhook mapping and the API transform. It takes a self-contained structural param (`{ draft?; work_in_progress? }`) so it stays assignable from `@gitbeaker`'s `MergeRequestSchema` without depending on that type declaring `work_in_progress`.
- New helper `isGitlabDraftToReadyTransition()` in the webhook handler: the Draft→Ready trigger now fires when **either** `changes.draft` **or** `changes.work_in_progress` goes `previous: true → current: false`.

## Assumption (please sanity-check)

The issue documents the MR **object** carrying `work_in_progress: true`, but does **not** prove the webhook `changes` payload carries `changes.work_in_progress: { previous, current }` on old GitLab. The added OR-branch is safe regardless: if that field never appears the branch never fires, and the return is idempotent (no double-trigger). Happy to adjust if maintainers know the real old-GitLab `changes` shape.

## Tests

Added unit tests following the repo's existing pattern of testing exported pure functions:

- `libs/common/utils/webhooks/gitlab.spec.ts` — `resolveGitlabIsDraft` truth table (including the `draft:null, work_in_progress:true` bug case) + a `mapPullRequest` integration case.
- Extended `libs/platform/infrastructure/webhooks/gitlab/gitlabPullRequest.handler.spec.ts` — `isGitlabDraftToReadyTransition` cases for both fields and the reverse (Ready→Draft) direction.

```
npx jest libs/common/utils/webhooks/gitlab.spec.ts
npx jest libs/platform/infrastructure/webhooks/gitlab/gitlabPullRequest.handler.spec.ts
```

> Note: I could not run the local jest suite in my environment (dependencies not installed), so I am relying on CI to run these. The pure-helper logic was verified directly. Please let CI validate.

### DIFF
--- a/libs/common/utils/webhooks/gitlab.spec.ts
+++ b/libs/common/utils/webhooks/gitlab.spec.ts
@@ -1,0 +1,57 @@
+import { GitlabMappedPlatform, resolveGitlabIsDraft } from './gitlab';
+
+describe('resolveGitlabIsDraft', () => {
+    it('treats draft:null + work_in_progress:true as draft (old GitLab, issue #1514)', () => {
+        expect(
+            resolveGitlabIsDraft({ draft: null, work_in_progress: true }),
+        ).toBe(true);
+    });
+
+    it('returns true when draft is true', () => {
+        expect(resolveGitlabIsDraft({ draft: true })).toBe(true);
+    });
+
+    it('lets an explicit draft:false win over work_in_progress:true', () => {
+        expect(
+            resolveGitlabIsDraft({ draft: false, work_in_progress: true }),
+        ).toBe(false);
+    });
+
+    it('falls back to work_in_progress:false when draft is null', () => {
+        expect(
+            resolveGitlabIsDraft({ draft: null, work_in_progress: false }),
+        ).toBe(false);
+    });
+
+    it('returns false when both fields are absent', () => {
+        expect(resolveGitlabIsDraft({})).toBe(false);
+    });
+
+    it.each([[null], [undefined]])('returns false for %p', (mr) => {
+        expect(resolveGitlabIsDraft(mr as any)).toBe(false);
+    });
+});
+
+describe('GitlabMappedPlatform.mapPullRequest — draft detection', () => {
+    it('maps a Draft MR reported with draft:null + work_in_progress:true as isDraft=true', () => {
+        const payload: any = {
+            object_kind: 'merge_request',
+            event_type: 'merge_request',
+            object_attributes: {
+                iid: 8423,
+                title: 'Draft: something',
+                state: 'opened',
+                draft: null,
+                work_in_progress: true,
+                source_branch: 'feature',
+                target_branch: 'main',
+                labels: [],
+            },
+            repository: { name: 'repo' },
+        };
+
+        const result = new GitlabMappedPlatform().mapPullRequest({ payload });
+
+        expect(result?.isDraft).toBe(true);
+    });
+});
--- a/libs/common/utils/webhooks/gitlab.ts
+++ b/libs/common/utils/webhooks/gitlab.ts
@@ -13,6 +13,24 @@ import {
 
 import { extractRepoFullName } from './webhooks.utils';
 
+/**
+ * Resolves whether a GitLab merge request is a draft.
+ *
+ * GitLab renamed "WIP" to "Draft" in 13.2. On older self-managed instances a
+ * Draft MR can be returned with `draft: null` and `work_in_progress: true`, so
+ * reading `draft` alone maps such MRs as non-draft. We treat `draft` as
+ * canonical when present (an explicit `draft: false` wins) and only fall back to
+ * the deprecated `work_in_progress` when `draft` is null/undefined.
+ */
+export function resolveGitlabIsDraft(
+    mr:
+        | { draft?: boolean | null; work_in_progress?: boolean | null }
+        | null
+        | undefined,
+): boolean {
+    return Boolean(mr?.draft ?? mr?.work_in_progress ?? false);
+}
+
 export class GitlabMappedPlatform implements IMappedPlatform {
     mapUsers(params: {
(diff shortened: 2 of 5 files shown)


vaibzde:
@kody review

Re-triggering the review — the previous run failed with an internal `openai_compatible` provider error before it could evaluate the diff, which looks like a transient issue on the review service side rather than something in this PR.

Wellington01:
@vaibzde this is already on main — #1516 landed the same fix. `libs/common/utils/webhooks/gitlab-draft.utils.ts` has `resolveGitlabDraftStatus` (`draft ?? work_in_progress ?? false`) and `isGitlabDraftToReadyChange`, wired into the same three sites you identified: `gitlab.ts:77`, the `transformPullRequest` in `gitlab.service.ts`, and `gitlabPullRequest.handler.ts`. It ships with its own spec.

The four conflicts on this branch are all in those files, for that reason. Nothing left to apply — suggest closing.

Answering the assumption you flagged: main's `isGitlabDraftToReadyChange` only falls back to `changes.work_in_progress` when `changes.draft` does not carry a complete `previous`/`current` boolean pair. A complete draft change that is not a Draft→Ready flip returns false instead of consulting the deprecated field.

PR #1652, opened 2026-08-04
feat(llm): support Application Default Credentials for Vertex AI

### LINKED ISSUE #1745: Support Application Default Credentials for Vertex AI (keyless auth)
## Problem

Running Vertex AI requires a service-account key in `API_VERTEX_AI_API_KEY` (base64 SA JSON). On a cluster with Workload Identity available, that means keeping static, long-lived key material on disk purely to satisfy the SDK — the one credential the platform is designed to let you avoid.

## Request

Support Application Default Credentials as an alternative to an explicit key. Omitting `googleAuthOptions` lets `google-auth-library` resolve credentials itself, which covers:

- GKE / EKS Workload Identity
- attached service accounts on GCE / Cloud Run
- a local `gcloud auth application-default login`

The project id can't come from the credentials in this mode — a workload identity credential is an `external_account` file with no `project_id` — so it needs to come from the environment (`GOOGLE_CLOUD_PROJECT`, with `GCLOUD_PROJECT` as an alias).

## Suggested shape

Keyless is opt-in and never changes existing behaviour: an explicit `API_VERTEX_AI_API_KEY` always wins, and ADC is only attempted when no key is set *and* a project is configured. An explicitly configured `API_OPEN_AI_API_KEY` should also take precedence, since `GOOGLE_CLOUD_PROJECT` is set by default on GCE/Cloud Run and shouldn't silently switch a deployment to Vertex.

Whatever selects the provider has to be applied consistently across every site that mirrors it — `byokToVercelModel`, `getInternalModel`, and `getModelName` — or a keyless deployment routes correctly for main calls while the internal model resolves to null and reporting names the wrong provider.

## Scope

The v5 (Vercel AI SDK) path in `libs/`. The v2 langchain adapters in `packages/kodus-common` need the same treatment but ship as a separately published package, so that would be a follow-up.

cursedcoder:
Closes #1745

## Problem

Vertex AI currently requires a service-account key. Every path parses a credentials string, requires `project_id` inside it, and hands the result to google-auth-library as explicit credentials:

- `libs/llm/byok-to-vercel.ts` — `if (!credentials?.project_id) return null`, then `googleAuthOptions: { credentials }`
- `packages/kodus-common/.../vertexAnthropicAdapter.ts` — `new GoogleAuth({ credentials })`

So there's no way to run Vertex without long-lived key material, and deployments on GKE/EKS can't use Workload Identity even when the platform already hands the pod an identity.

You can't work around it by passing a workload-identity credential either: that file is `type: "external_account"` and has no `project_id`, so it fails the guard and no model is built.

## Change

Adds a keyless path. Omitting `googleAuthOptions` lets google-auth-library resolve credentials itself, which covers Workload Identity, attached service accounts on GCE/Cloud Run, and a local `gcloud auth application-default login`. Since the project can no longer come from the key, it's read from `GOOGLE_CLOUD_PROJECT` (or `GCLOUD_PROJECT`) — the variables google-auth-library reads anyway.

**Opt-in and backwards compatible.** ADC is attempted only when *no key is configured* **and** a project is set in the environment:

| `API_VERTEX_AI_API_KEY` | `GOOGLE_CLOUD_PROJECT` | Behaviour |
|---|---|---|
| SA JSON | anything | unchanged — key wins, project from the key |
| not SA JSON (e.g. `AIzaSy…`) | anything | unchanged — AI Studio fallback |
| empty | set | **new** — ADC |
| empty | unset | unchanged |

Existing deployments all have a key, so nothing changes for them.

Applied at all three Vertex call sites in `libs/llm/byok-to-vercel.ts` (env-mode Gemini, env-mode Claude, BYOK `GOOGLE_VERTEX`), plus docs in `.env.schema` / `.env.example` and 6 unit tests.

## Scope

This covers the v5 (Vercel AI SDK) path in `libs/`. The v2 langchain adapters in `packages/kodus-common` need the same treatment, but that ships as a separately published package (`@kodus/kodus-common` is pinned at `1.3.20`, not a workspace link), so it needs its own release. Happy to send that as a follow-up if you'd like it.

Also left alone: `test-byok-connection.use-case.ts`. "Test these credentials" is a bit undefined when there are no credentials to test — tell me how you'd want it to behave and I'll add it.

## Testing

CI is green on the unit suite: `PASS libs/llm/byok-to-vercel.spec.ts`.

The tests extend the existing `byokToVercelModel — Google Vertex protocol routing` suite and reuse its mocking style, asserting that `googleAuthOptions` is absent, that `GCLOUD_PROJECT` works as an alias, that an explicit key still wins, that nothing happens with no project configured, and — after review feedback — that `getInternalModel` no longer returns null on an ADC deployment and that `getModelName` reports the provider actually serving the request.

I still can't run the suite locally: my network intercepts TLS to registry.npmjs.org, so `pnpm install` can't fetch tarballs (`UNABLE_TO_GET_ISSUER_CERT_LOCALLY`). Correction to an earlier version of this description — I claimed `pnpm@11.9.0` was unpublished and that the `packageManager` pin was a bug. That was wrong: the pin is fine, and my registry view was unreliable for the same TLS reason. Thanks to @Wellington01 for catching it.

The two remaining integration failures (`create-or-update-config-race`, `relation "organizations" does not exist`) are a test-harness DB issue on this branch's base, not something this change can reach — that spec passes on current `main`. Rebasing.

## Context

We run self-hosted Kodus on EKS with Workload Identity available, and wanted to drop the static SA key from the cluster. The infrastructure side is ready; this is the application-side piece that makes it usable.


cursedcoder:
Thanks — one of the three is a real bug and is now fixed in `ad715dc`. Notes on all three, plus the CI state.

### 3. ADC ordering vs `API_OPEN_AI_API_KEY` — real, fixed

The precedence concern is right and I've gated the ADC attempt on `!openaiKey`, mirroring the claude branch where the native Anthropic key already wins. Added a test for it.

The stated mechanism doesn't hold, though, and it's worth recording why: routing gemini through an OpenAI-compatible gateway requires `API_OPENAI_FORCE_BASE_URL`, and `isProxyBaseURL` returns `true` for *any* baseURL that isn't `api.anthropic.com` — so `viaProxy` is true and the whole `isGemini && !viaProxy` block is skipped. That configuration never reaches the ADC code.

What is reachable: `gemini-*` + no Google key + `API_OPEN_AI_API_KEY` set + no baseURL override. Previously that fell through to `createOpenAICompatible` at `api.openai.com` (which would also fail, since OpenAI doesn't serve `gemini-*` — but fails with the user's own key, predictably). The guard preserves that exact behaviour, so it's worth having regardless.

### 2. Unlogged catch — not applicable here

`byok-to-vercel.ts` is a module of pure functions, not a Nest provider: there's no `this` and no injected `PinoLoggerService`, so `this.logger.error(...)` can't be used as written. The sibling `vertexModelFromSaJson` — which this function is deliberately modelled on — uses the identical bare `catch { return null; }`, and both are "try this auth path, else let the caller fall through", where null is the meaningful signal rather than an error.

Happy to thread a logger through if you want it, but it'd be a change to the existing function too, so it felt out of scope for this PR rather than making the two siblings inconsistent.

### 1. No linked issue

There isn't an existing issue — I hit this while trying to drop a static SA key from a self-hosted deployment on EKS. Glad to open one to link if that's your preference for feature PRs.

### CI

`libs/llm/byok-to-vercel.spec.ts` **passes** on the first run. The failures don't look related to this change:

- `test/integration/parameters/create-or-update-config-race.integration.spec.ts` fails with `QueryFailedError: relation "organizations" does not exist` — migrations don't appear to have run against the test Postgres. This PR touches no DB code.
- `eligibility` (2s), `Comment preview URLs` (5s), `Deploy web preview` (26s) and `check` (36s) all fail almost immediately, which is the usual shape of a fork PR not having access to secrets.

Also flagging separately from the review: `packageManager` pins `pnpm@11.9.0`, which isn't published (npm's latest is `10.15.1`), so `pnpm install` fails at the self-install step for anyone outside your environment. That plus TLS interception on my network is why I couldn't run the suite locally and leaned on CI.

Wellington01:
@cursedcoder two sites in `byok-to-vercel.ts` mirror this provider selection and were not updated.

`getInternalModel` (line 630) — its own comment states it exists to match `byokToVercelModel`'s provider selection. Both Vertex branches are still gated on `vertexKey` (lines 663 and 679), so a keyless ADC deployment skips both, has no `openaiKey` either, and reaches `return null` at line 697. That null propagates: `resolveSecondaryPassModel` returns it (`secondary-pass-model.ts:56`) and `agent-review.stage.ts:1894` treats null as "skip the pass". So on an ADC deployment Vertex serves the main review calls while the secondary pass — dedup, severity classification, formatting — is silently skipped. No error, just degraded reviews.

`getModelName` (line 583) — same key-gated selection, used for reporting. An ADC deployment would report a different model than the one actually serving the request.

Both need `vertexModelFromAdc` / `vertexProjectFromEnv` applied.

On the description: `pnpm@11.9.0` is published (latest is 11.22.0), so the `packageManager` pin is not a bug — your registry view was stale or hit by the same TLS interception that broke the tarball fetches.

PR #1659, opened 2026-08-06
fix(cli): surface contextual message for non-JSON 403 responses

### LINKED ISSUE #1503: CLI: bare `Request failed with status 403` on larger `kodus review` runs (no actionable error)
## Summary

`kodus review` fails with a bare `Request failed with status 403` — printed right after `- Checking authentication...` — with **no actionable detail** (no mention of quota, rate limit, payload size, permissions, or which request failed). The same authenticated team key succeeds on smaller reviews in the same session, so the 403 is not a login problem, yet the message gives the user nothing to act on.

## Environment

- `@kodus/cli` **0.4.19**
- macOS (darwin), Node 24.6
- Auth: Team Key (`kodus auth status` / `team-status` both report ✓ Authenticated)

## What I observed (same valid team key, same session)

| Review | Result |
|---|---|
| `kodus review --prompt-only <one-file>` (1 file diff) | ✅ works |
| `kodus review --prompt-only <6 files>` (explicit) | ✅ works (`FILES_ANALYZED: 6`) |
| `kodus review --prompt-only` with **37** changed files (default working scope) | ❌ `Request failed with status 403` |
| `kodus review --prompt-only --branch main` run **inside a `git worktree`** (37-file diff) | ❌ `Request failed with status 403` |

Earlier in the same session, `--branch main` reviews with many files **did** succeed; larger reviews started returning 403 later. Small (1–6 file) reviews keep working throughout.

## Two concrete asks

1. **Surface a descriptive error instead of a bare `Request failed with status 403`.** Whatever the cause (usage/quota cap, rate limit, request-size/file-count limit, or a per-request permission), the CLI should print *why* and *what to do* (e.g. "review quota exceeded, resets at …", or "diff too large: N files > limit M — narrow the scope"). Right now the user cannot tell a transient limit from a real failure.

2. **If there is a per-review size / file-count limit**, please document it and fail with that message client-side before submitting, rather than surfacing the server's 403. If instead this is git-context detection breaking inside a `git worktree` (where `.git` is a file, not a directory), that path may need handling too — I could not fully isolate worktree-vs-size because a worktree's committed changes can only be reviewed via `--branch`, which is exactly the large-diff case.

## Repro sketch

```bash
# small review — works
kodus review --prompt-only path/to/one-file.ts

# large review (many changed files) — 403 with no detail
kodus review --prompt-only            # working tree with ~37 changed files
# or, inside a git worktree:
kodus review --prompt-only --branch main
```

Not blocking (small-batch reviews are a workable path), but the opaque 403 cost a lot of time to diagnose.

andyst-dev:
Fixes #1503

## Problem

`kodus review` fails with a bare `Request failed with status 403` on larger runs — no mention of quota, rate limit, payload size, or which endpoint failed. The user cannot tell a transient limit from a real failure.

**Root cause:** when the API returns a 403 with a non-JSON body (gateway/CDN page, empty body), the CLI synthesized its own `{ message: 'Request failed with status 403' }` payload. Because that string is ASCII, `normalizeApiErrorMessage` returned it verbatim (the 403 branch prefers a clean server message over the fallback), masking the actionable per-endpoint message that already exists: `Access denied for Kodus API endpoint (<path>).`

## Change

`apps/cli/src/services/api/api-core.ts` (both request paths):
- Only build the error payload from the JSON body.
- For non-JSON error responses, pass an empty payload so `normalizeApiErrorMessage` falls back to the contextual per-status message (`Access denied for Kodus API endpoint (<path>).` for 403, plus the existing rate-limit / unavailable / not-found messages for other statuses).

## Validation

- Added regression test: a 403 with `Content-Type: text/plain` now surfaces `Access denied for Kodus API endpoint (/cli/config/repositories/available).` instead of the raw placeholder.
- `api.real.test.ts`: 27/27 pass (26 existing + 1 new); `imports.test.ts`: 6/6.

### DIFF
--- a/apps/cli/src/services/api/__tests__/api.real.test.ts
+++ b/apps/cli/src/services/api/__tests__/api.real.test.ts
@@ -611,6 +611,31 @@ describe('RealApi config repository methods', () => {
         );
     });
 
+    it('surfaces the contextual endpoint message when a 403 has no JSON body', async () => {
+        // Non-JSON error bodies (gateway/CDN pages, empty bodies) used to
+        // surface the raw "Request failed with status 403" — a synthesized
+        // placeholder that masked the actionable per-endpoint message.
+        fetchMock.mockResolvedValue(
+            new Response('Forbidden', {
+                status: 403,
+                headers: { 'Content-Type': 'text/plain' },
+            }),
+        );
+
+        const api = new RealApi();
+
+        await expect(
+            api.config.getAvailableRepositories('kodus_team_key'),
+        ).rejects.toEqual(
+            expect.objectContaining({
+                name: 'ApiError',
+                statusCode: 403,
+                message:
+                    'Access denied for Kodus API endpoint (/cli/config/repositories/available).',
+            } satisfies Partial<ApiError>),
+        );
+    });
+
     it('sends X-Team-Key when getting repository settings', async () => {
         fetchMock.mockResolvedValue(
             new Response(
--- a/apps/cli/src/services/api/api-core.ts
+++ b/apps/cli/src/services/api/api-core.ts
@@ -351,7 +351,7 @@ export async function request<T>(
     if (!response.ok) {
         const rawError = isJson
             ? await response.json().catch(() => ({ message: 'Request failed' }))
-            : { message: `Request failed with status ${response.status}` };
+            : {};
         const errorData: ApiErrorPayload =
             rawError &&
             typeof rawError === 'object' &&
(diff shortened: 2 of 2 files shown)


Wellington01:
@andyst-dev this one is already on main — #1660 landed the same fix on 11 Aug (`fd4b6d48b`), for the same issue. Both call sites in `api-core.ts` already pass `{}` for non-JSON error bodies, and `api.real.test.ts` already carries the same 403 regression test with the same expected string.

So the conflict isn't worth rebasing — there's nothing left to apply. Suggest closing this one.

The change itself was right, for the record: with an empty payload the normalizer short-circuits to `getDefaultApiErrorMessage`, and for any status without a contextual branch that returns `Request failed with status N` — the exact string the synthesized payload used to produce, so no other status regressed.

PR #1664, opened 2026-08-08
fix(e2e): make the matrix report what it actually verified

### LINKED ISSUE #1699: command-review on Bitbucket: duplicate @kody delivery leaves only a 'skipped' execution row
### Symptom

`command-review × self-hosted × bitbucket × license-paid` fails with:

```
Review of PR #510 completed UNHEALTHY: execution status stayed "skipped"
through the full 90s window with no success row
```

Runs [31531667234](https://github.com/kodustech/kodus-ai/actions/runs/31531667234) and [31534832505](https://github.com/kodustech/kodus-ai/actions/runs/31534832505). It passed on retry in [31504796937](https://github.com/kodustech/kodus-ai/actions/runs/31504796937), which is what a timing-dependent race looks like. Other providers are unaffected.

### What the worker log shows

`@kody start command detected in Bitbucket comment` appears **twice**, and both deliveries end in:

```
[AutomationCodeReviewService] Code review already being processed for PR#510, skipping
[PipelineExecutor] Pipeline 'CodeReviewPipeline' skipped due to SKIP status
```

So Bitbucket delivers the comment webhook more than once and the product dedupes — which is the right behaviour on its own. The problem is what the E2E can see afterwards: within the 90s health window, the only TERMINAL execution row for that PR is the `skipped` duplicate.

`findExecutionStatus` already prefers `success` over `skipped` when both exist (there is a comment about exactly this case, observed earlier on the QA GitLab tenant), so this is not the assertion picking the wrong row when a good one is present. There simply was no successful row inside the window.

### The open question

Whether the first review settles late or never settles at all. Two very different answers:

- **Settles late** → the 90s health window is too short for Bitbucket, whose whole cell already runs 27-53 min against ~20 for the others. Test-side fix.
- **Never settles** → a review the user explicitly asked for is silently dropped, with no comment on the PR and nothing telling them why. Product bug, and a user-visible one.

Deciding needs the execution rows for the PR at failure time. The evidence bundle does not capture them today — worth adding regardless of the outcome here, since "what did the executions table actually say" is the first question every time this shape appears.

### Meanwhile

`command-review×bitbucket×license-paid` is quarantined in `tests/e2e/lib/verdict.ts`: the cell still runs and reports, but does not gate. Remove the entry when this closes.

malinosqui:
## Why

The matrix reported **green on runs where entire cells never executed**.

The 2026-08-07 release gate: 84 results = 23 passed, 2 failed, **5 skipped for GitHub quota**, 1 for a missing secret, 53 not applicable. The `github×free` and `github×community-byok` columns ran **zero** scenarios. The reported verdict was "2 failures".

Nothing in the pipeline could distinguish *we checked and it passed* from *we never checked*. That ambiguity is upstream of all three complaints — bugs escaping, flakiness, no visibility.

## What changed

**Typed skips + a third verdict.** `skipped` collapsed four unrelated outcomes into one number. It now carries a `SkipKind`: `not-applicable` (expected), `setup` (missing secret — a coverage gap), `infra` (quota / target down — **not verified**). An `infra` skip on a P0 scenario makes the run **INCONCLUSIVE** — exit 3, job red, distinct Discord message. Headlines read against executed/applicable, not total.

**Flaky is counted.** A pass that only happened on the retry is flagged and counted separately. The bare `/timeout/i` retry pattern is gone — it also matched deterministic mismatches, buying a wrong value a second LLM review.

**Durable history.** Results only lived in artifacts expiring in 14–30 days, so "when did this cell last pass?" had no answer — how a broken github cell stayed broken for six days. Each run appends JSONL to the `e2e-history` orphan branch, from a separate job so `contents: write` never touches the job holding provider tokens. This is CI data and stays **out of the product database**, whose migrations ship to every self-hosted install.

`pnpm --dir tests/e2e run report` → last-green per cell, flake rate, executed coverage. A weekly workflow posts it to Discord. `history-backfill.sh` seeds from surviving artifacts.

**GitHub quota.** GitHub signals rate limits with **403, not 429**, so the existing `Retry-After` handling never fired for it and the runner dropped the whole cell. Short resets (secondary limits, ~60s) are now waited out, capped at 5 min; longer ones surface and skip honestly. `currentUserId()` resolves from the durable PAT — the only reason self-hosted github cells couldn't use the App token. They now can, moving their heavy traffic onto the App's own 5000/h budget.

**Decoupled gates.** The self-hosted matrix had no cron; its only automatic caller was the release train, behind the cloud gate. On 07-31 and 08-07 the cloud gate failed and self-hosted validation was skipped entirely. It now runs Tue/Thu on its own, resolving the newest published release tag. Nightly cloud coverage returns to `fast.yml` (cut to 3 cells on 07-29 after 15 red nights — that made runs green by checking ~3% of what they used to); `vars.E2E_CLOUD_NIGHTLY_MATRIX` dials it back without a code change.

**Comment placement.** Scenarios asserted a review *arrived*, never *where it landed* — the shape of the GitLab "comments only on added lines" bug. `lib/diff-position.ts` validates anchors against the diff; `code-review-basic` asserts it, self-skipping on providers that don't report anchors.

## Testing

- 127 unit tests pass (40 new), typecheck clean
- `run-matrix --dry-run` on `fast.yml` and `full.yml`
- `run → history-append → report` end to end, text and markdown output
- All workflow YAML parses; shell scripts pass `bash -n`

Live provider scenarios can't be exercised from a dev machine. The placement assertion's logic is fully unit-tested and its live wiring self-skips.

## Not in this PR

Browser scenarios for the web flows (`apps/web` ties for the most fix churn in the last 10 weeks and has almost no matrix coverage) and a concurrency scenario for the pool-exhaustion class. Both need decisions about fixture repos and can't be validated from here.

## Deploy notes

- The `e2e-history` branch is created automatically on the first run.
- Run `tests/e2e/provisioning/history-backfill.sh` once to seed history from artifacts that still exist (oldest surviving is 08-01).
- Nightly cloud cost returns to pre-07-29 levels. Set `vars.E2E_CLOUD_NIGHTLY_MATRIX=matrix/midweek-smoke.yml` to keep it small.

---

Closes #1699

Related (open, tracked separately — this PR makes each one visible rather than fixing it):
- #1686 — onboarding backfill: gate + cap for providers with no repo-level comment listing
- #1687 — self-hosted infers the LLM provider from the model name
- #1695 — GitHub App path has no matrix coverage
- #1696 — scheduled matrix goes red on unreleased routes (RBAC manifest drift)
- #1700 — `@kody review` silently dropped while a PR automation holds the lock


malinosqui:
@kody review

PR #1679, opened 2026-08-10
fix(byok): map reasoning effort per Anthropic model generation

sartorijr92:
## Problem

Picking **Low / Medium / High** on an Anthropic BYOK config produced a request the provider rejected. Users worked around it by pasting raw JSON into the **Custom** field — which is why this started as "the Custom option doesn't open its textarea".

Anthropic changed the thinking API twice, and each generation returns a hard 400 on the others' shape:

| Generation | Models | Shape |
|---|---|---|
| legacy | 3.x, 4.0/4.1, Sonnet 4.5, Opus 4.5, Haiku 4.5 | `thinking.budget_tokens` (required) |
| 4.6 | Opus 4.6, Sonnet 4.6 | `thinking.type=adaptive` + `effort` |
| 4.7+ | Opus 4.7/4.8, Opus 5, Sonnet 5, Fable 5 | same, and `budget_tokens` is a 400 |

Two independent defects produced the same symptom:

1. The generation check only matched `claude-*-4-[6-9]`, so `claude-opus-5`, `claude-sonnet-5` and `claude-fable-5` fell through to the legacy branch.
2. **The code-review loop never passed a `modelName`.** The check ran against `undefined`, so *every* Anthropic model took the legacy path — not just the 5-series.

## Approach

Resolve the generation from the model id (`libs/llm/anthropic-model-traits.ts`) and pass the model name down the review path.

Only the **older** generations are enumerated — that list is closed. Anything newer matches the open-ended patterns, so a Claude released after this merges resolves to the modern shape with no code change.

### Behavior changes worth reviewing

- **Unidentifiable model → omit the thinking config** instead of guessing. The wrong guess is a 400 that kills the review; omitting only costs thinking depth. This inverts the previous default (and one existing test asserted the old behavior — updated, with the reasoning in the test body).
- **"Off" now sends `thinking.type=disabled` explicitly.** Opus 5 and Sonnet 5 think by default, so omitting the config billed thinking to users who picked Off. Fable/Mythos reject `disabled`, so thinking stays on there.
- **`temperature` withheld from 4.7+**, which removed sampling params. Scoped to the real `anthropic` provider — `anthropic_compatible` endpoints (Kimi Code, Z.ai, DeepSeek) still accept it, and `kimi-k2.7-code` *requires* `temperature=1`. The BYOK form hides the field and clears the stored value rather than letting it 400 at review time.

## Verification

Live providers, checked against the `[thinking] providerOptions resolved` log:

| Scenario | Payload sent | |
|---|---|---|
| haiku-4.5 / medium | `enabled` + `budgetTokens: 15000` | ✅ |
| sonnet-4.6 / low | `adaptive` + `effort: low` | ✅ |
| sonnet-5 / medium | `adaptive` + `effort: medium` | ✅ |
| opus-5 / low | `adaptive` + `effort: low` | ✅ |
| opus-5 / off | `disabled` | ✅ |
| kimi-k2.6 via anthropic_compatible / off | `{}` | ✅ |
| kimi-k2.6 via anthropic_compatible / medium + temp 1 | `budgetTokens`, temperature preserved | ✅ |

Plus 549 unit tests across the touched suites, and the front-end matrix driven in a real browser (Haiku 4.5 and Opus 4.6 show the temperature field; Opus 4.8, Opus 5 and provider-without-model hide it; OpenAI unaffected).

## Out of scope

Kody Rules fails on the Kimi Code plan with `tool_choice 'required' is incompatible with thinking enabled`. Reproduced **identically with thinking on and off**, so the thinking comes from the subscription plan's model, not from this mapping — a pre-existing incompatibility between the kody-rules agent and any provider that forces thinking. Filed separately.

The first commit (`fix(byok): make the Custom reasoning-effort textarea appear on click`) is the React Compiler bug that started this: it memoizes `watch("name")`, freezing the value at first render so the Custom textarea never appeared. Already reviewed on the previous branch, carried over here.


sartorijr92:
@kody review --force

sartorijr92:
@kody review

sartorijr92:
@kody review

PR #1680, opened 2026-08-10
fix(crons): throttle pool-starving cron bursts + supporting indexes

Wellington01:


Wellington01:
@kody start-review

Wellington01:
@kody start-review

PR #1681, opened 2026-08-10
ci: validate the api DI graph on every PR

malinosqui:
## Problem

A missing provider is a **runtime** failure raised while Nest builds the container, so nothing in CI can currently see it:

- `nest build` and `tsc --noEmit` only compile. A constructor parameter whose provider is not registered in that module is perfectly valid TypeScript.
- The unit suites instantiate classes directly or build purpose-made testing modules. None of them assemble the real `ApiModule` graph.

The miss that motivated this: a class is registered by **two** modules, a PR adds a constructor dependency, and only one of the two modules gets the matching `imports:` entry. Build green, tests green, container dead on boot. The grep that exposes the asymmetry only runs if somebody remembers to run it — DI resolution shouldn't depend on that.

## How it works

`scripts/ci/validate-di-graph.ts` builds the real graph:

```ts
await NestFactory.createApplicationContext(ApiModule, {
    preview: true,
    abortOnError: false,
});
```

`preview: true` walks the graph and resolves every constructor's dependencies **without instantiating a single provider** — the guard sits inside Nest's `instantiateClass`, while `resolveConstructorParams` runs unguarded. `abortOnError: false` surfaces the full `available in the Y module` text instead of a truncated log, which is the part that names what you forgot.

The script distinguishes the two failure shapes and prints the right remediation, since they are opposites: a missing provider needs an added `imports:` entry, while an `UnknownExportException` needs the module re-exported rather than the token.

## Verification

Both failure shapes and the green path were exercised end-to-end through the real CI entrypoint (`pnpm run ci:validate-di-graph`):

| Check | Result |
| --- | --- |
| Green on `main` | exit 0, graph resolves in ~115ms |
| Missing provider | exit 1, names the class, the argument and the owning module |
| `UnknownExportException` | exit 1, names the token and the module |
| Opens no DB/Redis/RabbitMQ/HTTP | 0 real network handles after the build; only stdio fds 0/1/2 remain |
| Runtime | ~10s total, install-bound; graph build itself ~150ms |

The no-I/O claim was checked two ways: by inspecting `process._getActiveHandles()` after the build, and by a run with every DB/broker host pointed at the black-hole address `192.0.2.1`, which still passes in the same time — a real connection attempt would block.

## Cost

Needs no secrets and no populated `.env`. Four throwaway placeholders satisfy the import-time guards and the Joi schema that run before the graph is built: `API_CRYPTO_KEY` and `CODE_MANAGEMENT_SECRET` (parsed as 32-byte hex at module scope), `CODE_MANAGEMENT_WEBHOOK_TOKEN` (asserted non-empty), and `API_PORT` (the only Joi-required key). No DB or broker host is set at all.

Scoped to `ApiModule` deliberately. `ReportCliModule`, `BackfillModule` and `mcp-manager` are follow-ups once this is proven.

## Note for reviewers

Test-drive against `feat/kodus-trace` (#1667) surfaced two **real** boot-breaking DI bugs on that branch's HEAD, both of which this gate catches and neither of which is fixed here:

1. `libs/code-review/pipeline/code-review-pipeline.module.ts:216` exports `LOAD_EXTERNAL_CONTEXT_STAGE_TOKEN`, a token `PromptsModule` owns. Nest only permits exporting a token from your own `providers:`, or re-exporting the whole module class.
2. `libs/code-review/pipeline/stages/finish-comments.stage.ts:42` — `UpdateCommentsAndGenerateSummaryStage` injects `PostTracePrCommentUseCase`, but `CodeReviewPipelineModule` never imports `TraceContextModule`.

### DIFF
--- a/.github/workflows/di-graph-gate.yml
+++ b/.github/workflows/di-graph-gate.yml
@@ -1,0 +1,86 @@
+name: DI Graph Gate (api)
+
+# Builds the real ApiModule dependency graph and fails the PR if Nest cannot
+# resolve it. This is the class of bug nothing else in CI can see:
+#
+#   - `nest build` and `tsc --noEmit` only compile. A constructor parameter
+#     whose provider is not registered in that module is valid TypeScript.
+#   - The unit suites instantiate classes directly or build purpose-made
+#     testing modules; none of them assemble the real ApiModule graph.
+#
+# The miss that motivated this gate: a class registered by TWO modules gains a
+# constructor dependency, and only one of the two modules gets the matching
+# `imports:` entry. Build green, tests green, container dead on boot.
+#
+# Nest's `preview: true` walks the graph and resolves every constructor's
+# dependencies without instantiating a single provider, so this opens no
+# database, Redis, RabbitMQ or HTTP connection and needs no real credentials.
+# See scripts/ci/validate-di-graph.ts.
+
+on:
+    pull_request:
+        paths:
+            - "apps/api/**"
+            - "libs/**"
+            - "packages/**"
+            - "scripts/ci/validate-di-graph.ts"
+            - ".github/workflows/di-graph-gate.yml"
+            - "tsconfig*.json"
+            - "package.json"
+            - "pnpm-lock.yaml"
+
+permissions:
+    contents: read
+
+jobs:
+    api-di-graph:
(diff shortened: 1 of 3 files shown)


malinosqui:
@kody review

malinosqui:
@kody review

Issue #256, opened 2025-09-20
Can't connect self hosted GitHub app


LivioGama:
With Github app I'm getting a nextauth error. I don't see any pkce at any moment of my flow

<img width="1125" height="276" alt="Image" src="https://github.com/user-attachments/assets/9b8d8299-203b-45cf-9387-29801a157a48" />

maybe it's related to setImmediate not existing at deploy:

<img width="856" height="324" alt="Image" src="https://github.com/user-attachments/assets/c9c8376a-6a8d-47b1-be4d-be55f3da8ab6" />

LivioGama:
[Add_`checks`_to_OAuth_providers_and_disable_`useSecureCookies`.patch](https://github.com/user-attachments/files/22440259/Add_.checks._to_OAuth_providers_and_disable_.useSecureCookies.patch)

I think this is the fix

malinosqui:
@Wellington01 can you help us here, please?

Wellington01:
Hi @LivioGama, how are you?

That’s strange, could you send me the link for that correction? Today we try to keep the packages as up to date as possible to ensure compatibility with the features. We are not encountering this kind of issue currently in our environments.

Could you clarify a few things:
	•	Are you trying to sign up on the platform using GitHub, or are you trying to connect to GitHub?
	•	Are you following the instructions in this documentation and using this project?
	         - https://docs.kodus.io/how_to_deploy/en/deploy_kodus/generic_vm
	         - https://github.com/kodustech/kodus-installer

Issue #1689, opened 2026-08-11
Support Azure DevOps as a task-management MCP for business-logic review


mzglinski:
### Is your proposal related to a problem?

Self-hosted teams that use Azure Boards / Azure DevOps for work items can connect a custom Azure DevOps MCP plugin, but Kodus still shows **“Connect a task management MCP”** and skips business-logic review.

Today task-management detection only recognizes managed/hinted providers (`jira`, `linear`, `notion`, `clickup`, `googledocs`, `atlassianrovo`, `githubissues`). A plugin named **Azure DevOps** (provider `custom`) is ignored.

Even when forced past that gate, Kodus does not understand common Azure Boards naming:

- PR titles like `feat(AB#5625): …`
- Branches like `ORG-5625`
- Work-item URLs on `dev.azure.com` / `visualstudio.com`

So business-logic review cannot reliably fetch the linked Azure work item via the connected MCP (`get_work_item`, `search_work_items`, etc.).

### Describe the solution you'd like

Treat Azure DevOps / Azure Boards as a first-class task-management MCP:

1. Recognize plugin names/providers containing `azuredevops` / `azureboards` (and skill examples including **Azure DevOps**) so the UI badge clears and the business-logic stage/agent preflight accept the connection.
2. Detect Azure Boards signals from PR title, branch, and body:
   - `AB#<id>` (emit both `AB#5625` and bare `5625` for MCP `get_work_item`)
   - Azure work-item URLs (`/_workitems/edit/<id>`, `/workitems/<id>`)
   - Existing `PROJ-123`-style keys when present (e.g. branch `ORG-5625`)
3. Add Azure DevOps URL domain patterns (`dev.azure.com`, `visualstudio.com`) alongside the other task MCP URL matchers.
4. Keep GitHub `#N` / PR-number extraction separate so Azure work items are not confused with GitHub issue/PR numbers.

### Describe alternatives you've considered

- **Rename hack** (e.g. put `jira` in the plugin display name): can clear the UI gate via hint matching, but does not teach Kodus `AB#` / Azure URLs, and steers heuristics toward the wrong tracker.
- **Rename to GitHub Issues**: worse fit — enables `#N` extraction and risks treating GitHub PR numbers (e.g. `/pull/74`) as task IDs.
- **Ask developers to only paste Jira-style keys or foreign tracker URLs**: fights existing Azure Boards conventions (`AB#`, Azure work-item links).
[...]

PR #1690, opened 2026-08-11
feat: support Azure DevOps as a task-management MCP

mzglinski:
## Summary

Implements #1689: recognize Azure DevOps / Azure Boards as a task-management MCP for business-logic review, including `AB#` work-item refs and Azure DevOps work-item URLs, so custom Azure DevOps MCP plugins clear the UI gate and can supply task context.

This PR covers **repo on GitHub, boards in Azure DevOps**. Native Azure Repos `getIssue` / `listIssues` (Git Issues plugin for Azure-hosted repos) is a follow-up, not this change.

## Test plan

- [x] Unit tests for `BusinessLogicValidationStage` (`AB#` extraction, Azure work-item URLs, Azure DevOps MCP skip gate) — passing
- [x] `ORG-5625`-style keys when the connection matches through the `azureboards` hint
- [x] Full unit suite on push (`SKIP_INTEGRATION=true` where local Postgres unavailable)
- [ ] Connect a custom plugin named `Azure DevOps` and confirm the business-logic settings badge no longer asks to connect a task-management MCP
- [ ] Run business-logic review on a PR titled `feat(AB#…)` / branched `PROJ-123` and confirm the agent receives work-item signals (not GitHub PR `#N`)

## Notes for the reviewer

- Closes / implements https://github.com/kodustech/kodus-ai/issues/1689 (custom-MCP / `AB#` path only)
- Touch points: UI `TASK_MANAGEMENT_HINTS`, business-logic stage hints/URL patterns/`AB#` detection, skill examples, task-reference helpers
- Intentionally does **not** treat GitHub `/pull/N` as an Azure work item
- Hint matching: `normalizeProviderToken` lowercases and strips non-alphanumerics; aliases checked are `provider`, `name`, `metadata.connection.{id,serverName,appName}`. So **Azure DevOps** and **Azure Boards** match; **ADO** and **Azure** do not.


Wellington01:
@mzglinski three things.

1. `detectTicketKeys` now emits `AB#N` and the bare `N` unconditionally, but `hasRelevantBusinessSignals` gates ticket keys on `TICKET_KEY_MCPS` — any of jira, linear, notion, clickup, githubissues, gitissues, atlassianrovo. So a PR titled `feat(AB#5625)` in an org with only Jira connected now returns true at `business-logic-validation.stage.ts:635` and the stage proceeds, handing the agent `["AB#5625", "5625"]` to resolve against Jira. Before this change `AB#5625` produced no ticket keys at all (`TICKET_KEY_PATTERN` needs `ABC-123`, and the `#N` branch requires start-of-string or whitespace/paren before the `#`), so that PR skipped with `no_signals`. The dedicated Azure branch at line 658 is correctly gated on `azuredevops`/`azureboards`, but it is unreachable in that case because line 635 already returned. Either gate the `AB#` push in `detectTicketKeys` on a connected Azure MCP, or drop `AB#` keys from the ticket-key path and rely on the branch at 658.

2. Same problem via the URL path: the work-item regex is not domain-restricted, unlike `MCP_URL_PATTERNS` which limits Azure to `dev.azure.com`/`visualstudio.com`. `https://anything.example.com/workitems/123` produces ticket keys and takes the same route through line 635.

3. The `AZURE_BOARDS_REF_REGEX` loop added to `extractIssueNumbers` (`task-references.ts`) is a no-op. `ISSUE_NUMBER_REGEX` already matches the `#5625` inside `AB#5625` — its `#` alternative has no left boundary — and both loops add to the same `Set<number>`. Verified against `feat(AB#5625): x`, `AB#12 and AB#34`, `Fixes AB#7`: identical output with and without the new loop.

CI is red on this branch (backend tests, web preview, eligibility) — worth a rebase and a re-run, the last one is from a while ago.

mzglinski:
@Wellington01 

1 and 2 - this is the same coupling GitHub Issues already has. `detectTicketKeys` emits `#N` (and ids from any `/issues/N` URL, not only github.com) with no MCP check, then `hasRelevantBusinessSignals` treats any ticket key as enough if any `TICKET_KEY_MCPS` connection is present. So `Closes #256` in a Jira-only org already proceeds, and the dedicated `gitissues`/`githubissues` branch is skipped the same way you describe for Azure. `AB#` / work-item URLs follow that existing pattern rather than introducing a new one. I am not going to special-case Azure keys or Azure URL hosts in this PR; a real fix belongs in a follow-up that changes how all non-Jira keys (GitHub `#N`, Azure `AB#`, unrestricted `/issues/` and `/workitems/` URLs) interact with the skip gate. Happy to help with that refactor if you want it, but it should be a separate PR.

3 - agreed, the `AZURE_BOARDS_REF_REGEX` loop in `extractIssueNumbers` is a no-op (`ISSUE_NUMBER_REGEX` already matches the `#N` inside `AB#N`). Dropping it.

CI - preview and eligibility tests fail because I do not have enough permissions in this repo, and the backend tests fail at Postgres-related stuff, because it requires migrations that do not run on PRs from forked repositories.

Wellington01:
@mzglinski point 3 is resolved, thanks. Traced the full path end to end before merging — the runtime chain works, but only for a custom MCP whose connection name normalizes to `azuredevops` / `azureboards`. Three things.

1. `azureboards` went into `TASK_MANAGEMENT_HINTS` and `MCP_URL_PATTERNS` but not into `TICKET_KEY_MCPS` (`business-logic-validation.stage.ts:64`) — only `azuredevops` was added there. A connection matched through the `azureboards` hint gets `AB#5625` accepted by the new branch at line 656, but a `PROJ-123` / `ORG-5625` key is rejected at the ticket-key branch. #1689 lists that case explicitly ("Existing `PROJ-123`-style keys when present (e.g. branch `ORG-5625`)"). Either add `azureboards` to `TICKET_KEY_MCPS` or drop it from the other two and keep one canonical hint.

2. Nothing here reaches `apps/mcp-manager/src/config/managed-mcp-servers.json`, so Azure DevOps never appears in Settings → Plugins and `category` is never stamped on the connection. Category is the *primary* match in both gates — `serverMatchesRequiredCategory` (`generic-skill-runner.service.ts:544`) and `conn.category === 'task-management'` in `getConnectedTaskManagementMcps` — with the name hint only as fallback. Without a catalog entry the feature rides entirely on the customer typing the right connection name.

3. When that catalog entry lands, `MANAGED_TASK_MCP_HINT` (`business-logic-validation.stage.ts:54`) needs the id → canonical hint mapping as well. The comment above it already spells out the failure mode: falling back to the raw id "would never match `TICKET_KEY_MCPS` / `MCP_URL_PATTERNS`".

Worth adding to the PR description, since it is not discoverable: `normalizeProviderToken` lowercases and strips non-alphanumerics, and the aliases checked are `provider`, `name`, `metadata.connection.{id,serverName,appName}`. So "Azure DevOps" and "Azure Boards" match, "ADO" and "Azure" don't.

Tool discovery is fine — `task.context.read` is `provider_dynamic`, so `get_work_item` / `search_work_items` resolve from the connected server with no allowlist change.

On CI: the only red suite is `create-or-update-config-race.integration.spec.ts` (`relation "organizations" does not exist`), which fails on `main` and on internal PRs too — not a fork-permission thing, and not yours. 630 suites / 5905 tests pass.

Wellington01:
Splitting my previous comment by owner, so it's clear what's blocking this PR and what isn't.

**Yours — one line, in a file you already touch.** `libs/code-review/pipeline/stages/business-logic-validation.stage.ts`, `TICKET_KEY_MCPS` (line 64): add `'azureboards'` next to `'azuredevops'`, or drop `'azureboards'` from `TASK_MANAGEMENT_HINTS` and `MCP_URL_PATTERNS` and keep a single canonical hint. Right now the three lists disagree and `ORG-5625`-style keys are rejected on a connection matched through the `azureboards` hint. That's the only change I'd block a merge on.

**Ours, not yours — don't start on it.** Points 2 and 3 (the managed catalog entry and `MANAGED_TASK_MCP_HINT`) can't be done from this PR. Azure DevOps ships an official remote MCP server, GA since 2026-08-05, but its endpoint is per-organization — `https://mcp.dev.azure.com/{organization}` — and every entry in `apps/mcp-manager/src/config/managed-mcp-servers.json` has a static `baseUrl` with no templating or user-supplied URL segment. Adding Azure DevOps as a managed plugin therefore needs a schema change in `apps/mcp-manager`, which is a separately released service. We'll take that.

So this PR stays scoped to what it already does: the custom-MCP path. With the one-line fix above it's consistent, and it's the correct first step regardless of how the managed entry lands later.

Wellington01:
Retracting my point 2 from above — the managed-catalog entry is not what's missing, and it isn't blocking anything here. Details in https://github.com/kodustech/kodus-ai/issues/1689#issuecomment-5360267399, short version:

The **Git Issues** plugin already covers task-management for every host, without any external MCP: its `baseUrl` is `/mcp/issues` (we host it), its auth is `type: "none"`, and it reads issues through the code-management integration the team already connected. What gates it is `CodeManagementService.isIssuesSupported()`, which returns false when the platform adapter has no `listIssues`. GitHub, GitLab, Bitbucket and Forgejo implement `getIssue`/`listIssues`; `azureRepos.service.ts` does not. Adding them there — mapped onto Azure Boards work items, same credentials we already use for PRs — is the real fix for teams whose repos live in Azure DevOps. That's ours, tracked in the issue.

So this PR keeps a scope of its own and is worth merging as-is: it covers **repo on GitHub, boards in Azure DevOps**, where `AB#` is the Microsoft-recommended way to cross the two systems and there is no native link to read. The adapter work doesn't replace it.

Still the one thing I'd like changed before merge, unchanged from my first comment: `azureboards` is in `TASK_MANAGEMENT_HINTS` and `MCP_URL_PATTERNS` but not in `TICKET_KEY_MCPS` (`business-logic-validation.stage.ts:64`), so `ORG-5625`-style keys are rejected on a connection matched through that hint. Add it there, or drop `azureboards` and keep `azuredevops` as the single canonical hint.

Wellington01:
@mzglinski withdrawing the "ours, not yours" from my earlier comment — if you want the adapter work, it's yours.

Everything needed is written up in https://github.com/kodustech/kodus-ai/issues/1689#issuecomment-5360267399: the two method signatures, the provider-agnostic return shape, the full field mapping from an Azure Boards work item, the two caveats that will bite otherwise (`System.State` varies by process template, so map on state *category*; `number` is org-wide in Azure, not per-repo), and the REST endpoints including the WIQL two-step.

I'd suggest a separate PR for it rather than growing this one — this one is ready once `azureboards` goes into `TICKET_KEY_MCPS`, and there's no reason to hold it behind a different subsystem. But if you'd rather do it all here, that's fine too.

mzglinski:
@Wellington01 added `azureboards` to `TICKET_KEY_MCPS` (with a regression test for `ORG-5625` when only that hint is connected). Also restored the em-dash in the hash test name and dropped the no-op `extractIssueNumbers` AB# case.

Leaving the Azure Repos `getIssue` / `listIssues` adapter for a separate PR — this one stays on the GitHub-repo / Azure Boards custom-MCP path.

Connection names that normalize to `azuredevops` / `azureboards` match (`Azure DevOps`, `Azure Boards`); `ADO` and `Azure` do not.

Sebdevar (approved):

mzglinski:
@Wellington01 Is there anything else you want me to address in this PR before it can be merged?

jairo-litman:
@mzglinski I ran this end to end against a real Azure DevOps org instead of just reading the diff. The diff does what it says, but the path it opens doesn't work, so it can't merge as-is.

## Setup

Scratch Azure DevOps org, boards only. A work item `AB#8` with a full description and acceptance criteria. A GitHub PR titled `feat(AB#8): …` that implements it only partially, so a working business-logic review had something real to flag. The official Azure DevOps remote MCP (`https://mcp.dev.azure.com/{org}`) connected as a custom plugin named `Azure DevOps`, all 42 tools enabled. Two review runs.

## What works

Both unchecked boxes in your test plan, twice:

```
[BUSINESS-LOGIC] running business-rules agent (passed skip gate)
```

Badge clears, no `no_task_mcp` / `no_signals` skip, `AB#8` extracted from the title and handed to the agent. Unit suites pass on top of current `main` (63/63) — the branch needs `main` merged first or stale `kodus-common` breaks the imports.

## 1. The hint lists still disagree

Add `Azure Boards` to `examples` in `libs/agents/skills/business-rules-validation/SKILL.md`, or drop `azureboards` and keep `azuredevops` as the single hint.

`resolveRequiredProviderHints` builds the skill preflight's hints from that `examples` string, and `providerMatchesRequiredHints` won't match `azureboards` against `azuredevops` in either direction. Custom MCPs have `category: null`, so the name hint is the only path. Tested against the real service:

- named `Azure DevOps` → preflight resolves
- named `Azure Boards` → `RequiredMcpPreflightError`

That name gets a cleared badge, a stage that runs, then a review that dies at preflight. It also means the `TICKET_KEY_MCPS` fix from last round **can't be exercised end to end** — `azureboards` only reaches `TICKET_KEY_MCPS` through a connection matched by that hint, which is the connection preflight rejects. Runs pass today only because `Azure DevOps` is in both lists.

## 2. The work item is never fetched

Neither run called `wit_work_item` or `search_workitem`. Both are in `allowedTools`; both appear in the registry listing the agent gets.

| | Run 1 | Run 2 |
|---|---|---|
| Tool calls | 426 across 5 tools | 42, all `search_wiki` |
| Work-item calls | **0** | **0** |
| Outcome | 300s timeout | `task_context_status:weak` |

Run 1 put the ticket key into whatever string parameter came first:

```
repo_repository failed: Unknown action 'AB#8'. Supported: 'get', 'list'.
wiki_upsert_page failed: The 'etag' value must be … Received: AB#8
```

Probable cause: Azure's MCP is action-dispatch shaped — `wit_work_item` takes `action: 'get' | 'get_batch' | …` rather than being a plain getter — and with 42 tools the selection never lands on it. Fetching the same work item by hand through that server works fine (`wit_work_item`, `action: "get"`, id 8).

What a customer gets instead:

> 🤔 **Insufficient Task Context** — I found a task linked to this PR, but it only contains minimal information (title only, no description or acceptance criteria).

That's false, and it's the end state of the feature this PR ships: Kody telling an Azure Boards user to write a description for a ticket that already has one.

## 3. A read-only capability is handed the write surface

`task.context.read` is `provider_dynamic` with no `tools` list (`libs/agents/skills/skill-capabilities.ts:30`), so the lookup gets everything the server exposes. Run 1 made **330 write attempts** — `wiki_upsert_page`, `repo_pull_request_thread`, `testplan_*_write`; run 2 made none. Nothing landed (wikis still `count: 0`, work item still `rev: 1`), but they failed on Azure's parameter validation, not on anything of ours, and it varies run to run.

Jira and Linear expose small read-shaped surfaces, which is why this has never come up. This PR is the first supported way to attach a large write-capable MCP to that capability, so it arrives with this change.

## Direction

2 and 3 likely have one fix: give `task.context.read` a scoped tool set instead of `provider_dynamic` with no list. A per-provider read allowlist (`wit_work_item`, `search_workitem` for Azure; the equivalents elsewhere) cuts 42 tools down to the two that matter — which should fix the selection — and removes the write surface at the same time. If you'd rather keep discovery dynamic, a read-only filter plus explicit hinting toward work-item tools would also work, but it leaves the tool-choice problem to the model.

The bar I'd want before merge is one live run on an `AB#` PR where the agent returns the actual work-item description, not `task_context_status:weak`. Happy to re-run my setup against your branch to check.

Two fixes for the PR description while you're in there:

- The tools are `wit_work_item` and `search_workitem`, not `get_work_item` / `search_work_items`. No runtime impact since the capability is `provider_dynamic`, but #1689 repeats it.
- Auth deserves a sentence. Bearer tokens are stored statically and never refresh — only `OAUTH2` does (`oauth.ts:334`) — and OAuth against Azure DevOps can't use dynamic registration, because the Entra tenant's discovery document has no `registration_endpoint`. A customer has to create an app registration and supply clientId/secret.

mzglinski:
@jairo-litman I'm considering closing this PR, as my organization is debating switching task management software, which means we will soon need a different solution anyway. 

Since this PR only addresses a specific use case for a custom task management MCP, would you be open to exploring a more flexible direction? 

My rough idea is to add a setting to the custom plugin configuration that allows users to flag it as a task management MCP. In this config, users could provide a list of regex patterns to extract task numbers from the PR title or description. I think this approach would make the integration universal for any task management software. 

If you think this is a good path forward, I’m happy to create a feature request and a design doc so we can iterate on the details before moving to implementation.

jairo-litman:
@mzglinski Yeah, I like that better than what's here. Closing this is fine by me.

Talked it over with Wellington and we're basically where you are. Rough shape:

- a flag on the custom plugin config marking it as a task management plugin
- on that plugin, patterns the user defines for pulling the task id out of the PR title/description/branch
- the user picks which of that plugin's tools get used for task lookup
- a config modal on the business logic setting (like the Kody Rules one) where you choose which plugin provides task context

Then Jira, Linear, whatever comes next — point a custom plugin at their MCP with your own auth, mark it, set the pattern, pick the tools. Nothing hardcoded per provider.

The pattern part fixes something that's been biting us anyway. Right now if there's a `#8223` or some URL sitting in a PR body we have no idea whether it's a task or something unrelated, so we guess. If the user tells us, we stop guessing.

A few things I'd want nailed down in the doc:

Getting the id out is the easy part. Turning it into a tool call is where I'd spend the time. `wit_work_item` wants `action: 'get'` plus an id, Jira wants a key, everyone's different, and ticking tool checkboxes doesn't tell the agent how to call them. Cutting a server's 40-odd tools down to two or three might be enough on its own — worth actually trying that before we design around it either way.

If users pick the tools, they can pick write tools. Do we hard filter to read-only or just default to it?

Whatever we build can't break the managed plugins. Jira/Rovo and GitHub Issues go through the current hints and need to keep working — probably managed keeps its own behavior and custom uses the config.

And a raw regex box is a bit of a footgun. Presets for the common ones, something that shows you what matches as you type, and a guard so a bad pattern can't hang the pipeline.

Wellington also floated letting the plugin carry a bit of its own prompt context. Might be past v1 but worth a mention.

Go for it on the feature request and design doc, tag me or Wellington and we'll take a look.

mzglinski:
Closing this PR in favor of a generic task-management MCP design instead of Azure-only hint lists.

The skip-gate / `AB#` / name-hint approach does not fetch the work-item body and does not stop the fetcher from seeing write tools (as in @jairo-litman’s 42-tool Azure MCP run).

Canonical design: https://github.com/kodustech/kodus-ai/issues/1884 (spec in the follow-up comment).

cc @jairo-litman @Wellington01

PR #1714, opened 2026-08-13
fix(bitbucket): read Data Center webhook payload from pullRequest key

### LINKED ISSUE #1713: Bitbucket Data Center webhooks read the PR payload from the wrong key
### Issue Summary

Pull request webhooks coming from **Bitbucket Data Center / Server** are silently dropped and never produce a code review.

Bitbucket Data Center sends the pull request object under the top-level key `pullRequest` (capital R), as documented by Atlassian in the [Event payload reference](https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html).

`BitbucketMappedPlatform`, in `libs/common/utils/webhooks/bitbucket.ts`, reads it from `pullrequest` (all lowercase). Since JavaScript object keys are case sensitive, the lookup returns `undefined` and every mapper method bails out on its first guard.

Nothing normalizes the key in between: `apps/webhooks/src/controllers/bitbucket.controller.ts` spreads the raw body and only appends `isDataCenterEvent`.

Bitbucket **Cloud** is not affected, because Cloud genuinely sends `pullrequest` in lowercase. This only affects self-hosted Data Center / Server installations.

I believe this is a bug rather than intended behaviour for three reasons:

1. The Data Center branch of `mapPullRequest()` reads `toRef`, `fromRef` and `displayId`, which are fields that only exist in the Data Center payload. So the branch was written deliberately for Data Center, it just reads them from the wrong parent key.
2. `IWebhookBitbucketDataCenterPullRequestEvent` declares `eventKey` in camelCase, which is correct per Atlassian, but `pullrequest` in lowercase, which is not. The inconsistency inside the same interface suggests the shape was written by hand rather than from a captured payload, and because the type declares the lowercase key, TypeScript cannot catch the mismatch.
3. There is no test or fixture anywhere in the repository covering a Data Center event, and `bitbucketPullRequest.handler.ts` is the only one of the five PR handlers without a spec file.

### Steps to Reproduce

1. Take the `pr:opened` payload exactly as published in Atlassian's Event payload reference for Bitbucket Data Center.
2. Add `isDataCenterEvent: true`, which is what `BitbucketController` appends before enqueueing.
3. Call `BitbucketMappedPlatform.mapPullRequest()`, `.mapRepository()` and `.mapUsers()` with that payload.
4. Repeat with the same payload, changing only the key `pullRequest` to `pullrequest`.

I did this by bundling `libs/common/utils/webhooks/bitbucket.ts` as-is with esbuild and calling the class directly, so the code under test is the repository's own, unmodified.

I also traced the downstream path by reading `bitbucketPullRequest.handler.ts`: with `mapRepository()` returning `null`, `repository.id` falls back to `payload?.repository?.uuid`, which does not exist in the Data Center payload, so `webhookContextService.getContext()` finds no context and the handler returns early. I did not execute that part, only the mappers.

### Actual Results

With the payload as Atlassian actually sends it, all three mappers return `null`:

- `mapPullRequest()` returns `null`
- `mapRepository()` returns `null`
- `mapUsers()` returns `null`

No exception is thrown. The only trace left is an info-level log from BitbucketPullRequestHandler, which reads Processing Bitbucket 'pr:opened' event for PR #undefined (URL not found), because prId falls back to payload?.pullrequest?.id and is also undefined. After that, repository.id resolves to undefined, getContext() is called with the string "undefined", and the handler returns early. From the outside the integration looks healthy while no pull request is ever reviewed.

If it helps confirm this on your side: searching production logs for PR #undefined on Bitbucket events should surface it.

### Expected Results

The same payload should map correctly, as it does when only the key casing is changed:

- `mapPullRequest()` returns the mapped PR (`number: 1`)
- `mapRepository()` returns the mapped repository (`id: 84`, `name: repo-1`)
- `mapUsers()` returns the mapped users

A Data Center pull request should reach the code review flow exactly like a Cloud one.

### Technical details

- Not browser related. This is a backend webhook mapping issue, so there are no console logs or network requests from a browser to attach.
- Node.js v22.22.0
- Repository at commit `6c47d9668`
- Affected files:
  - `libs/common/utils/webhooks/bitbucket.ts` (`mapUsers`, `mapPullRequest`, `mapRepository`)
  - `libs/platform/domain/platformIntegrations/types/webhooks/webhooks-bitbucket.type.ts` (`IWebhookBitbucketDataCenterPullRequestEvent`)
  - `apps/webhooks/src/controllers/bitbucket.controller.ts` (no normalization between receiving and enqueueing)
  - `libs/platform/infrastructure/webhooks/bitbucket/bitbucketPullRequest.handler.ts` (consumes the null results)
- Affected events: `pr:opened`, `pr:modified`, `pr:merged`, `pr:declined`, `pr:comment:added`, `pr:reviewer:updated`

### Evidence

I ran the repository's own mapper against both payloads. Output:

```
===========================================================
  Running the Kodus mapper
===========================================================

  1) Payload as Atlassian sends it   (key "pullRequest")
    mapPullRequest() -> null   <<< BROKEN
    mapRepository()  -> null   <<< BROKEN
    mapUsers()       -> null   <<< BROKEN

  2) Payload as the code expects     (key "pullrequest")
    mapPullRequest() -> ok  (PR #1)
    mapRepository()  -> ok  (repo 84 / repo-1)
    mapUsers()       -> ok

===========================================================
```

The only difference between the two runs is the capitalisation of one letter in one key. I am happy to attach the reproduction script if that helps.

### Question before opening a PR

Are there active Bitbucket Data Center users today? If this is worth fixing, my proposal would be:

1. Normalize once in `bitbucket.controller.ts`, right next to the existing `isDataCenterEvent` check, rather than spreading `??` across the handler.
2. Correct `IWebhookBitbucketDataCenterPullRequestEvent` so the type reflects the real payload.
3. Add a mapper spec plus Data Center cases to the handler, mirroring the existing GitLab and Forgejo specs.

Happy to open that PR, and equally happy to be told I have misread something. I am new to this codebase.

MatheusMartinho:
## Summary

Pull request webhooks from Bitbucket Data Center / Server were silently dropped and never produced a code review. Data Center sends the pull request under `pullRequest` (capital R), while the code reads `pullrequest` (lowercase), so every mapper returned `null` and the handler returned early without logging an error.

This normalizes the key once in `BitbucketController`, before the payload is enqueued, and adds the test coverage that was missing for Data Center events.

Closes #1713

## Test plan

- [x] `pnpm run test -- "test/unit/webhooks/(controllers|mappers)/bitbucket"` → 2 suites, 30 tests passing
- [x] New `bitbucket-mapped-platform.spec.ts` covers Cloud and Data Center side by side, plus the guards that return `null` (10 tests)
- [x] New `Data Center events` block in the existing controller spec covers the normalization, all 6 `pr:*` events, the case where Data Center sends no pull request at all, and a Cloud payload carrying a stray `pullRequest` key
- [x] The 11 pre-existing Cloud tests in `bitbucket.controller.spec.ts` still pass unchanged, so the Cloud path is untouched
- [x] **Mutation check**: reverted only the controller change and re-ran. 7 tests failed (the normalization test and all 6 cases of the `it.each`). Restored the fix and all 30 passed again, confirming the new tests actually lock the behaviour rather than passing regardless.

## Notes for the reviewer

**Why the controller and not the mapper.** `libs/common/utils/webhooks/bitbucket.ts` reads `payload.pullrequest` in 26 places and `chatWithKodyFromGit.use-case.ts` in 8 more. Normalizing once at the entry point fixes all of them without touching any of them. Fixing it inside the mapper alone would have left `bitbucketPullRequest.handler.ts` still broken, since it reads the key directly.

**The original key is preserved.** The spread keeps `pullRequest` alongside the normalized `pullrequest`, so nothing that might depend on the raw shape breaks and the enqueued payload still matches what Bitbucket actually sent. Happy to drop it if you would rather keep the queue message minimal.

**Scope kept deliberately narrow.** No new environment variables, nothing under `ee/`, and `pnpm-lock.yaml` untouched.

**One thing I could not verify.** I do not know whether there are active Bitbucket Data Center users today, so I cannot say how much impact this has in practice. If it helps confirm it from your side, searching production logs for `PR #undefined` on Bitbucket events should surface any affected installations, since `prId` also resolves to `undefined` on this path.

**The type change is documentation only.** `IWebhookBitbucketDataCenterPullRequestEvent` still declares `pullrequest`, which is correct for everything downstream of the controller. I added a JSDoc explaining that the raw event uses `pullRequest` and where the normalization happens, so the next person does not hit the same trap.

I am new to this codebase, so please push back on anything that does not match how you would have done it.

### DIFF
--- a/apps/webhooks/src/controllers/bitbucket.controller.ts
+++ b/apps/webhooks/src/controllers/bitbucket.controller.ts
@@ -54,6 +54,14 @@ export class BitbucketController {
                     event,
                     payload: {
                         ...payload,
+                        // Bitbucket Data Center sends the pull request under
+                        // `pullRequest` (capital R), while Cloud sends `pullrequest`.
+                        // Normalize once here so the mappers and the handler can rely
+                        // on a single key.
+                        // https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html
+                        ...(isDataCenterEvent && payload?.pullRequest
+                            ? { pullrequest: payload.pullRequest }
+                            : {}),
                         isDataCenterEvent,
                     },
                 })
--- a/libs/platform/domain/platformIntegrations/types/webhooks/webhooks-bitbucket.type.ts
+++ b/libs/platform/domain/platformIntegrations/types/webhooks/webhooks-bitbucket.type.ts
@@ -204,6 +204,12 @@ export interface IWebhookBitbucketDataCenterPullRequestEvent {
     eventKey: string;
     date: string;
     actor: IWebhookBitbucketDataCenterActor;
+    /**
+     * Bitbucket Data Center sends this as `pullRequest` (capital R).
+     * `BitbucketController` normalizes it to `pullrequest` before enqueueing,
+     * so every consumer downstream sees the lowercase key.
+     * @see https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html
+     */
     pullrequest: IWebhookBitbucketDataCenterPullRequest;
     comment?: IWebhookBitbucketDataCenterComment;
     commentParentId?: number;
--- a/test/unit/webhooks/controllers/bitbucket.controller.spec.ts
+++ b/test/unit/webhooks/controllers/bitbucket.controller.spec.ts
@@ -126,6 +126,122 @@ describe('BitbucketController', () => {
         });
     });
 
+    describe('Data Center events', () => {
+        /**
+         * Bitbucket Data Center sends the pull request under `pullRequest`
+         * (capital R), while Cloud sends `pullrequest`.
+         * @see https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html
+         */
+        const dataCenterBody = {
+            eventKey: 'pr:opened',
+            pullRequest: {
+                id: 1,
+                title: 'a new file added',
+                toRef: {
+                    displayId: 'master',
+                    repository: { id: 84, name: 'repo-1' },
+                },
+            },
+        };
+
+        it('should normalize the Data Center "pullRequest" key to "pullrequest"', async () => {
+            mockRequest = {
+                headers: { 'x-event-key': 'pr:opened' },
+                body: dataCenterBody,
+            };
+
+            controller.handleWebhook(
+                mockRequest as Request,
+                mockResponse as Response,
+            );
+
+            expect(mockResponse.status).toHaveBeenCalledWith(HttpStatus.OK);
+            expect(mockResponse.send).toHaveBeenCalledWith('Webhook received');
+
+            await new Promise((resolve) => setImmediate(resolve));
+
+            const enqueued = enqueueWebhookUseCase.execute.mock.calls[0][0];
+
+            expect(enqueued.payload.pullrequest).toEqual(
+                dataCenterBody.pullRequest,
+            );
+            expect(enqueued.payload.isDataCenterEvent).toBe(true);
+        });
+
+        it.each([
+            'pr:opened',
+            'pr:modified',
+            'pr:reviewer:updated',
+            'pr:comment:added',
+            'pr:merged',
+            'pr:declined',
+        ])('should enqueue "%s" with the normalized key', async (event) => {
+            mockRequest = {
+                headers: { 'x-event-key': event },
+                body: { ...dataCenterBody, eventKey: event },
+            };
+
+            controller.handleWebhook(
+                mockRequest as Request,
+                mockResponse as Response,
+            );
+
+            await new Promise((resolve) => setImmediate(resolve));
+
+            expect(enqueueWebhookUseCase.execute).toHaveBeenCalledTimes(1);
+
+            const enqueued = enqueueWebhookUseCase.execute.mock.calls[0][0];
+
+            expect(enqueued.event).toBe(event);
+            expect(enqueued.payload.isDataCenterEvent).toBe(true);
+            expect(enqueued.payload.pullrequest).toBeDefined();
+        });
+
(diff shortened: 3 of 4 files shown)


Wellington01:
@MatheusMartinho checked this properly, including re-running your mutation claim. The diagnosis and the fix are right.

What I verified rather than took on faith:

- `apps/webhooks/src/controllers/bitbucket.controller.ts` is the only `@Controller('bitbucket')` in the repo, so normalizing at the entry point covers everything. Every reader of `payload.pullrequest` — the handler plus the spots in `chatWithKodyFromGit.use-case.ts` — consumes the enqueued payload, not the raw request.
- The mapper's Data Center branch already reads genuine DC shapes (`toRef`, `fromRef`, `displayId` in `libs/common/utils/webhooks/bitbucket.ts:59-80`), which backs your reading of it: DC support was implemented, only the entry key was wrong. Nothing downstream needed touching, so keeping this to the controller was the right call.
- `isDataCenterEvent = event.startsWith('pr:')` lines up exactly with the six DC events in the allowlist, and every Cloud event starts with `pullrequest:`, so the gate can't misfire.
- Your mutation check reproduces here: dropping the normalization fails exactly 7. I also ran a mutant you didn't — removing only the `isDataCenterEvent &&` guard — and it kills exactly one test, the Cloud passthrough one, so that guard test isn't vacuous either. 30/30 green, CI green.

One change before merge: drop the original key. Nothing reads `payload.pullRequest` anywhere outside your own tests (grepped `libs/` and `apps/`), so keeping it doubles the PR object in every Data Center queue message for no consumer. If something ever needs the raw shape, adding it back is a deliberate decision at that point.

PR #1733, opened 2026-08-15
fix(e2e): repair RC harness blockers

malinosqui:
## What changed

- install Playwright Chromium before browser-backed E2E scenarios
- authenticate the Cockpit API probe directly instead of sending a JWT through the cookie-only web proxy
- isolate the dedicated SSO topology on DigitalOcean and tear it down after CI
- fail Bitbucket cells early when stale tunnel webhooks cannot be deleted
- suppress expected 404 noise when the optional Trace branch does not exist
- prevent shell command substitution from backticks inside the VM heredoc

## Why

The RC run `31842820287` exposed deterministic harness failures that obscured the product signal and consumed the full Bitbucket timeout.

## Validation

- `npm test` in `tests/e2e`: 177 passed
- focused Trace reader Jest suite: 5 passed
- `actionlint` on the two changed workflows: passed
- `git diff --check`: passed
- `bash -n` on the VM provisioner: passed

The E2E typecheck remains blocked by the pre-existing untracked `tests/e2e/benchmark/tp-lost.ts` (`TS18046`). The repository pre-push env coverage hook is also blocked by unrelated untracked work referencing `API_CRON_LICENSE_INACTIVITY` and `RUN_ID`.

### DIFF
--- a/.github/workflows/e2e-cloud.yml
+++ b/.github/workflows/e2e-cloud.yml
@@ -104,6 +104,10 @@ jobs:
               working-directory: tests/e2e
               run: npm install --no-audit --no-fund
 
+            - name: Install Playwright Chromium
+              working-directory: tests/e2e
+              run: npx playwright install --with-deps chromium
+
             # Seed the cloud tenant registry. The runner resolves each cloud
             # cell's tenant by (provider, license) from ~/.kodus-dev/cloud-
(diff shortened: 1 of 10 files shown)


malinosqui:
@kody review

sartorijr92:
@kody review

PR #1756, opened 2026-08-18
fix(gitlab): build clone URL from path_with_namespace, not stored fullName

### LINKED ISSUE #1743: GitLab clone URL is built from stored fullName, breaking after renames and for display-form paths
## Summary

`GitlabService.getCloneParams` builds the clone URL from `repository.fullName`, which is not reliably the GitLab URL slug. Cloning then fails, and the failure looks like an auth problem rather than a bad path.

## Detail

`fullName` holds whatever was persisted when the repository was first selected. It diverges from the slug in two ways:

1. **Display form** — `My Group/My Project` gets encoded to `My%20Group/My%20Project` instead of resolving to `my-group/my-project`.
2. **Stale after a rename or group move** — the stored value keeps the old path, and if the project gained a parent subgroup that segment is missing entirely (`group/project` where it now lives at `parent/group/project`).

## Symptom

GitLab answers an unresolvable path with a 302 to `/users/sign_in`, so git reports:

```
fatal: unable to update url base from redirection:
  asked for: https://gitlab.example.com/My%20Group/My%20Project/info/refs?service=git-upload-pack
   redirect: https://gitlab.example.com/users/sign_in
```

This reads as an expired token or a missing permission grant, but the credential is fine. Every other call in `GitlabService` goes by numeric project id and is immune to renames, so the integration looks healthy while only cloning breaks.

Downstream it is quiet: `createSandboxWithRepo` burns its full retry budget, `CreateSandboxStage` logs `all retries exhausted, continuing without it`, and the review silently degrades to self-contained mode (no tools, diff-only) with no signal to the user.

## Expected

The clone URL should use the project's current `path_with_namespace`, resolved by numeric id so it survives renames and moves.

## Affects

Self-hosted GitLab, any project whose stored `fullName` is not the exact current slug — display names with spaces or capitals, and any project renamed or moved since it was first connected.

malinosqui:
Closes #1743
Re-submits the fix from #1735 (same diff) on a same-repo branch — see why below.

## Problem

`GitlabService.getCloneParams` builds the clone URL by URL-encoding `repository.fullName`. That value is whatever was persisted when the repository was first selected and is not reliably the GitLab URL slug:

1. **Display form** — `My Group/My Project` encodes to `My%20Group/My%20Project` instead of resolving to `my-group/my-project`.
2. **Stale after a rename or group move** — the stored value keeps the old path, silently dropping a parent subgroup added later.

GitLab answers the bad path with a 302 to `/users/sign_in`, so git reports `unable to update url base from redirection` — which reads like an auth failure but isn't. Every other call in `GitlabService` goes by numeric project id and stays healthy, so the integration looks fine while only cloning breaks. Downstream, `createSandboxWithRepo` burns its retry budget and the review silently degrades to self-contained mode (diff-only, no tools) with no signal to the user.

## Fix

Resolve the slug via `Projects.show(id)` and use `path_with_namespace`, falling back to the stored `fullName` if the lookup fails or the id isn't resolvable (CLI mode passes a placeholder `'0'` with a `fullName` already parsed from the local remote, which is authoritative — `hasResolvableProjectId` skips the round-trip there).

## Why this PR exists alongside #1735

**Root-cause diagnosis and the fix itself are [@cursedcoder](https://github.com/cursedcoder)'s (#1735, closes #1743).** That PR runs from a fork (`cursedcoder/kodus-ai`), and three required checks fail there with 403s regardless of the diff's correctness:

- `eligibility` — `gh api POST .../statuses` needs write scope
- `Comment preview URLs` — needs `issues:write`/`pull_requests:write`
- `Deploy web preview` — needs the `RAILWAY_TOKEN` secret

None of these are repo-issues, they're the standard read-only `GITHUB_TOKEN` GitHub Actions gives `pull_request` events from forks. Re-running doesn't change it. This branch carries the identical fix from inside the repo so it can pass full CI and actually be merged.

## Independent verification (TDD split, see commits)

1. `test(gitlab): reproduce the fullName clone-URL bug (#1743)` — adds only the two bug-reproducing tests against **unmodified** `main`. Confirmed red: `builds the URL from path_with_namespace when fullName holds the display form` and `picks up a subgroup the stored fullName predates` fail; the other 19 cases in the file pass.
2. `fix(gitlab): resolve clone URL from path_with_namespace, not stored fullName` — adds the fix. Confirmed green: 21/21 pass.

Verified twice — once against a reconstructed patch, once against `cursedcoder`'s actual commit (`b7863c3e6`, same SHA CI tested on #1735) checked out byte-for-byte — both runs identical.

## Tests

`gitlab.service.spec.ts` — 21 cases (8 new/restructured for this bug): display form, subgroup move, CLI placeholder-id skip, lookup-failure fallback, and existing self-hosted-protocol behavior.

### DIFF
--- a/libs/platform/infrastructure/adapters/services/gitlab.service.spec.ts
+++ b/libs/platform/infrastructure/adapters/services/gitlab.service.spec.ts
@@ -97,27 +97,166 @@ describe('GitlabService', () => {
         });
     });
 
-    it('does not duplicate the protocol when building clone params for self-hosted GitLab', async () => {
-        jest.spyOn(service as any, 'getAuthDetails').mockResolvedValue({
-            accessToken: 'oauth-token',
-            authMode: AuthMode.OAUTH,
-            host: 'https://gitlab.example.com/',
+    describe('getCloneParams', () => {
+        const mockAuthDetails = () =>
+            jest.spyOn(service as any, 'getAuthDetails').mockResolvedValue({
+                accessToken: 'oauth-token',
+                authMode: AuthMode.OAUTH,
+                host: 'https://gitlab.example.com/',
+            });
+
+        const mockProjectsShow = (show: jest.Mock) =>
+            mockedGitlab.mockReturnValue({ Projects: { show } });
+
+        it('does not duplicate the protocol when building clone params for self-hosted GitLab', async () => {
+            mockAuthDetails();
+            mockProjectsShow(
+                jest.fn().mockResolvedValue({
+                    path_with_namespace: 'group/repo',
+                }),
+            );
+
+            const cloneParams = await service.getCloneParams({
+                organizationAndTeamData,
+                repository: {
+                    id: 'repo-1',
+                    name: 'repo',
+                    fullName: 'group/repo',
+                    defaultBranch: 'main',
+                },
+            });
+
+            expect(cloneParams.url).toBe(
+                'https://gitlab.example.com/group/repo',
+            );
+            expect(cloneParams.auth).toMatchObject({
+                type: AuthMode.OAUTH,
+                token: 'oauth-token',
+            });
         });
 
-        const cloneParams = await service.getCloneParams({
-            organizationAndTeamData,
-            repository: {
-                id: 'repo-1',
-                name: 'repo',
-                fullName: 'group/repo',
-                defaultBranch: 'main',
+        // A stored fullName can hold the display form. Encoding it verbatim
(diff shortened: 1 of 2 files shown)


malinosqui:
The fix in this PR shipped via #1735 (merged) — this branch's diff against `main` is now empty. Closing as duplicate. Kody's caching suggestion here is being addressed separately in #1757.

PR #1771, opened 2026-08-21
ci(e2e): surface matrix advisories and setup skips on the run page

malinosqui:
## Why

The self-hosted matrix gate is binary by design — advisory failures and setup skips never fail a job — so the Actions run page shows **all-green even when scenarios failed** or planned coverage silently did not run. Concrete case, run [32444831872](https://github.com/kodustech/kodus-ai/actions/runs/32444831872) (today's RC validation): every job green, yet the evidence carried **6 advisory failures** across the 4 cells — including a P1 product failure (`centralized-config-sync` → `centralized:selectRepo` HTTP 400) — and **6 setup skips** (e.g. `cockpit-analytics` 403, Vertex BYOK blinded by GCP quota). None of that was visible without digging into raw job logs, and the Discord message said just *"matrix green"*.

## What

The runner already writes everything needed per cell (`result.json`, `notify.json`, `summary.md`) — it just never reached the run page.

- **`tests/e2e/cli/render-scoreboard.py`** (new, stdlib-only): renders the downloaded evidence as
  - a per-cell scoreboard + advisory/setup-skip/flaky/unverified sections on the run's **Summary tab**, with each cell's full scenario table in a collapsed drill-down;
  - `::warning::`/`::error::` **annotations**, so the run page itself shows a count next to the green check;
  - digest outputs (`digest`, `gating_count`, `advisory_count`, `setup_skipped`) via `GITHUB_OUTPUT`.
- **`e2e-self-hosted-matrix.yml`**: aggregate job checks out the script (sparse), renders the scoreboard, and exposes the digest as `workflow_call` outputs. The previously conditional checkout for the failure notify is now the unconditional first step.
- **`selfhosted-build-push.yml`**: the *RC validated* Discord notification now carries the digest in its title (e.g. `37/43 passed · 6 advisory · 6 setup-skips`) and calls out advisory failures / coverage gaps before someone approves the promote.

**Gating is unchanged** — the script never exits non-zero; green stays green, it just stops looking perfect when it isn't.

## Testing

- Ran the script locally against the real combined evidence artifact of run 32444831872: scoreboard table, advisory/setup sections, annotations and digest all match the numbers extracted manually from the raw logs; also exercised the no-`notify.json` fallback (failures degrade to unclassified advisories instead of vanishing).
- YAML validated; actionlint reports only pre-existing shellcheck infos in untouched steps.
- Pushed with `--no-verify`: the pre-push env-coverage failure is from unrelated untracked WIP files in the local tree (`API_CRON_LICENSE_INACTIVITY`, `RUN_ID`), not from this change.

### DIFF
--- a/.github/workflows/e2e-self-hosted-matrix.yml
+++ b/.github/workflows/e2e-self-hosted-matrix.yml
@@ -690,23 +690,31 @@ jobs:
                       exit 0
                   fi
                   npm install -g @anthropic-ai/claude-code >/dev/null 2>&1
-                  # Read-only investigation; Write is allowed solely so the
-                  # agent can produce the two contract files (the prompt pins
-                  # that, and the runner is ephemeral either way).
+                  # The evidence the agent reads (droplet logs, scenario
+                  # errors) is untrusted text, and the NEXT step executes a
+                  # script from this checkout with GH_TOKEN — so the agent is
+                  # STRICTLY read-only: Bash/Write/Edit are denied (deny rules
+                  # take precedence over every allow, including any settings
+                  # file in the checkout). Its report arrives on stdout; the
+                  # trusted extractor below materializes the contract files.
+                  # --strict-mcp-config: never load MCP servers from any
+                  # config that lands in the checkout — MCP tools would be
+                  # an unvetted side door around the deny list.
                   claude -p "$(cat tests/e2e/cli/investigate-prompt.md)" \
-                      --allowedTools "Read,Grep,Glob,Write" \
+                      --allowedTools "Read,Grep,Glob" \
+                      --disallowedTools "Bash,Write,Edit,MultiEdit,NotebookEdit,WebFetch,WebSearch,Task" \
+                      --strict-mcp-config \
                       --max-turns 60 \
                       --output-format text > investigation-agent.log 2>&1 || {
                       echo "::warning title=E2E investigation failed (advisory)::agent run failed — see investigation-agent.log artifact"
                   }
+                  python3 tests/e2e/cli/extract-investigation.py investigation-agent.log
                   if [ -f investigation.md ]; then
                       {
                           echo "## 🔎 Failure investigation (agent — verify before acting)"
                           echo ""
                           cat investigation.md
                       } >> "$GITHUB_STEP_SUMMARY" || true
-                  else
-                      echo "::warning title=E2E investigation (advisory)::agent produced no investigation.md"
                   fi
 
             - name: Upsert one issue per failure signature
--- a/.github/workflows/e2e-self-hosted-matrix.yml
+++ b/.github/workflows/e2e-self-hosted-matrix.yml
@@ -561,14 +561,17 @@ jobs:
             # with the scoreboard; this is analysis, labeled as such. Skips
             # silently when E2E_LLM_API_KEY isn't configured and never
             # fails the job.
+            # Same secret/vars/defaults as the cell jobs' product LLM
+            # (DeepSeek V4 Flash on Fireworks) — one vendor, one key, one
+            # variable to debug. No extra secret to provision.
             - name: LLM read of the evidence (advisory)
               id: llm-note
               if: always()
               continue-on-error: true
               env:
                   E2E_LLM_API_KEY: ${{ secrets.E2E_LLM_API_KEY }}
-                  E2E_LLM_BASE_URL: ${{ vars.E2E_LLM_BASE_URL || 'https://api.deepseek.com' }}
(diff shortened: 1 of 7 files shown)


malinosqui:
Second pass (b2ad694bc), from live feedback:

- **Glanceable verdict**: the summary now opens with a single headline — `🟡 E2E matrix: GREEN WITH CAVEATS — 6 advisory failure(s), 6 coverage gap(s)` — followed by the attention sections; the per-cell table moved below.
- **Amber Discord**: new `status`/`discord_color` outputs; the RC-validated embed turns amber and says *with caveats* whenever advisories or coverage gaps hide behind the green gate. Clean run stays green.
- **LLM read (optional)**: `llm-release-note.py` asks a cheap OpenAI-compatible model (default DeepSeek; `E2E_LLM_MODEL`/`E2E_LLM_BASE_URL` repo vars override) to classify each failure — infra/quota vs test-side vs product — and say what should block the promote. Advisory-only: skips without `E2E_LLM_API_KEY`, never fails the job, counts stay with the deterministic scoreboard. Live test against run 32444831872's evidence correctly called Vertex quota 'infra noise' and the `centralized-config-sync` 400 'product P1, investigate before promote'.

**To enable the LLM note**: add the `E2E_LLM_API_KEY` repo secret (DeepSeek key). Without it everything else works unchanged.

malinosqui:
Third layer (4bb009d08): **agent root-cause investigation** on failed runs.

- New `investigate` job — only fires when the aggregate carries advisory/gating failures. A read-only Claude Code agent follows each failure through the droplet logs + evidence into the repo, classifies it (product / test-side / infra) with a cited evidence trail, and proposes a fix. Report lands on the run summary.
- One GitHub issue per stable failure signature (label `e2e-investigation`): first occurrence opens it, recurrences comment + reopen — a repeating failure stops being rediscovered every run.
- **Auth is the Claude subscription** (no per-token API billing): generate with `claude setup-token` and add as the `CLAUDE_CODE_OAUTH_TOKEN` repo secret. Without the secret the job skips silently; it is `continue-on-error` and can never gate a release.
- Validated locally against run 32444831872's real evidence — the agent root-caused the `centralized-config-sync` HTTP 400 that three runs had reproduced: **test-side** — the consolidated `GH_INTEGRATION_TOKEN` (PR #1742) cannot manage webhooks on the `kodus-config-e2e` fixture repo, so GitHub's hooks-API 404 surfaces as the opaque 400. Vertex failures filed as known infra quota. Issue upsert tested against a stubbed `gh` (create / recurrence / reopen / intra-run dedupe / bad input).

malinosqui:
Kody review addressed in 0b2d15949:

- **Prompt injection via unscoped Write (critical) — fixed, with a stronger design than first attempted.** Live testing confirmed the reviewer's concern and more: allowlist path-scoping (`Write(investigation.md)`) did NOT contain the agent — ambient allow rules leak through, and the agent routed writes via Bash. The design now inverts: the agent is strictly read-only (`--disallowedTools Bash,Write,Edit,...` — deny precedence beats every allow), `--strict-mcp-config` closes MCP side doors from the checkout, and the report arrives on stdout; the trusted `extract-investigation.py` materializes the contract files. Verified live: write attempts via Write and shell are both refused, and a full re-run on real evidence reproduced the same root causes through the stdout contract.
- **200-issue reconciliation cap — fixed.** Limit raised to 1000 (gh paginates internally) plus an explicit warning if the label ever reaches the cap, instead of silently duplicating issues.
- **Token-usage accounting rule (2×) — partially applied, rest not applicable.** `runAiSdkLLMInSpan` is NestJS-runtime infrastructure; a CI workflow step has no product runtime or org context to report into. The spirit is honored: the Fireworks call now logs provider-reported prompt/completion/total tokens to the job log, and the Claude investigation runs on subscription auth (no per-token billing).

malinosqui:
Second Kody pass addressed in 6effa00bb:

- **Extractor consistency bug — fixed.** Item selection and report truncation are now tied to the same fenced block (the last one parsing as a list of dicts with `signature`), so a malformed final block can no longer pair an earlier narrative array with a mismatched cut. Defense-in-depth: the upsert's signature validation already rejects placeholder/non-slug values, so even a slipped-through example block cannot create issues. Tested against an echoed-schema fixture, a malformed-final-block fixture, and the real agent output.
- **Token-usage accounting (repeat) — requesting the rule's own exception path.** `runAiSdkLLMInSpan` requires the NestJS runtime and an org context; neither exists in a CI workflow step, and building an adapter service for a test-tooling script would be scope inversion. Per the rule text ('or get an explicit maintainer-approved exception'), flagging this for maintainer sign-off: provider-reported token usage is logged on every call, and the investigation agent runs on subscription auth. @sartorijr92 to approve/dismiss.

PR #1807, opened 2026-08-29
docs: update PR file limit from 200 to 2,000

### LINKED ISSUE #1795: docs: update PR file limit from 200 to 2,000
## Problem

The public pricing documentation still says that Kodus does not review pull requests with more than 200 changed files:

- docs/how_to_use/en/pricing.mdx
- docs/how_to_use/pt-br/pricing.mdx

The current code review flow uses a limit of 2,000 analyzable files after removed files and configured ignore paths are excluded. This makes the public documentation inaccurate and can lead to incorrect customer comparisons.

## Expected change

- Update the English and Portuguese pricing pages from 200 to 2,000 analyzable files after ignores.
- Search the remaining public documentation for other references to the old 200-file limit and update them.
- Keep the wording scoped to the technical file limit. Large pull requests can still affect review cost, duration, and coverage.

## Implementation reference

- libs/code-review/pipeline/stages/fetch-changed-files.stage.ts defines AGENT_MAX_FILES as 2,000.
- FetchChangedFilesStage passes that value to validateFiles in the current flow.

Source: https://github.com/kodustech/kodus-ai/blob/main/libs/code-review/pipeline/stages/fetch-changed-files.stage.ts

MatheusMartinho:
Fixes #1795

## Summary

The docs still say Kody skips PRs with more than 200 changed files, but the code enforces 2,000. `AGENT_MAX_FILES` is set to `2000` in `libs/code-review/pipeline/stages/fetch-changed-files.stage.ts` and is applied to the file list after removed files and `ignorePaths` are filtered out, which is exactly the behaviour the troubleshooting page describes. This updates every customer facing mention so the docs match the code.

Files changed:

- `pricing.mdx` in all five locales (en, es, ja, pt-br, zh)
- `code_review/troubleshooting.mdx` in all five locales
- `docs/PR_CHEATSHEET.md`

In the troubleshooting page the worked example was scaled with the limit, so it still illustrates what ignore patterns do: 2,500 total files, 600 of them `.lock`, 1,900 reviewed. Number formatting follows each locale (2.000 for es and pt-br, 2,000 for en, ja and zh).

The wording stays scoped to the technical file limit, as the issue asks. Nothing was added about review cost, duration or coverage.

## Test plan

- [x] Grepped the whole repo for the old limit and confirmed the 11 files above are the only customer facing occurrences
- [x] Checked the constant in `fetch-changed-files.stage.ts` and where it is applied in the pipeline, to confirm 2,000 is the value actually enforced and that it counts files after filtering
- [x] Read the rendered diff of all 11 files, including the CJK and Japanese lines, to confirm no surrounding text was touched

## Notes for the reviewer

Two things worth a look:

1. The `zh` locale had both strings and was not in the list on the issue, so I included it. Without it the Chinese docs would be the only ones left saying 200.
2. `docs/PLANO_MELHORIAS_DOCS.md` also mentions 200, but it reads as an internal planning doc rather than customer facing, so I left it untouched. Happy to update it too if you prefer it in sync.

I also left alone the "200 to 400 lines" references in `pr_size.mdx` and `lead_time_for_changes.mdx`, since those are about PR size in lines and unrelated to this limit.

Disclosure: I used an AI assistant to sweep the locales for every occurrence. I verified the constant in the source, reviewed the final diff line by line and take responsibility for the change.


MatheusMartinho:
The one red check here, `Release freeze eligibility / eligibility`, is not coming from this diff. It fails on every pull request opened from a fork, and it would fail the same way on an empty commit.

What happens:

- `release-freeze-check.yml` runs on `pull_request`. When that event comes from a fork, GitHub hands the job a read-only `GITHUB_TOKEN` regardless of the `permissions:` block, so the `statuses: write` request is not honoured.
- The step then calls `gh api -X POST repos/${REPO}/statuses/${SHA}` and gets `gh: Resource not accessible by integration (HTTP 403)`. That is the whole log of the failing run on this PR.
- Because the script runs under `set -euo pipefail`, the 403 exits the job with code 1 after about 4 seconds.

The side effect is worth more than the red X: the `release-freeze-eligible` commit status is never posted at all for fork PRs, not even the quiet `success` the workflow intends to post while the freeze is off. So if the `release-freeze` ruleset requires that status, an outside contribution cannot satisfy it on its own.

It looks systemic rather than specific to this PR. Of the 1,088 runs of this workflow, 161 are failures, and the ones I sampled are all from fork branches.

A fix that would work: move the trigger to `pull_request_target`. This job never checks out the PR branch, it only reads `github.event.pull_request.title` and `.head.sha` from the event payload, so it does not carry the usual `pull_request_target` risk, and it would run with a token that can post the status. Worth noting the change only takes effect once it is on `main`, since `pull_request_target` always runs the workflow file from the base branch.

I did not want to push a change to a CI workflow from the outside without asking first. Happy to open that PR if you want it, or to leave it with you.

Either way, the docs change in this PR is unaffected and Kody has approved it.

PR #1819, opened 2026-09-01
fix(code-review): searchDocs says backend unavailable instead of faking no docs

### LINKED ISSUE #1762: searchDocs (Exa) tool is dead in production: returns "No documentation found" without ever calling Exa
## Summary

The finder/verifier agents' `searchDocs` VERIFY tool (backed by `DocumentationSearchExaService`) has **never returned a single real result in production** in the observable window, and it fails in a way that actively misleads the agent: it returns `No documentation found for "<pkg>" with query "<q>"` — semantically "I searched and there is nothing" — regardless of whether the Exa client was never constructed OR every Exa call threw and was swallowed. The agent cannot tell "tool unavailable/broken" from "no docs exist", so every third-party-API claim falls back to the model's prior.

This is a root cause of a confirmed false positive on PR #1759 (a `--testPathPatterns` Jest flag flagged as invalid at ERROR severity), and almost certainly of many others: the tool that exists precisely to verify third-party-API claims is silently non-functional.

## Evidence (Langfuse, prod project)

- Trace `4dfecf6279485fe55f2554d7daf23ff9` (`kodus-generalist-review-agent`, PR 1759, 2026-08-19 12:54 UTC): 5 `searchDocs` calls, all about Jest, all with the CORRECT hypothesis (e.g. `"v30 CLI option --testPathPattern renamed to --testPathPatterns"`), all returned `No documentation found for "jest"`.
- Last 24h, all orgs: **795 `searchDocs` TOOL observations; 100/100 sampled returned "No documentation found"**; production AND homolog; zero real doc hits.
- Sampled one day per week back to 2026-06-24 (oldest data present): **zero** calls with a real result. The only non-"No documentation found" outputs were 5× `Error: both packageName and query are required.` on 2026-08-05. As far as the data goes, this has never worked in prod.
- Latency of these calls is ~90–150 ms — the same order as `grep`/`readFile` (which run in the remote sandbox), so latency does **not** discriminate between the two causes below. (An earlier draft of this issue claimed "sub-millisecond, never left the process" — that was a unit-reading error on my part; retracted.)

## Two candidate causes — identical symptom, distinguishable ONLY in worker logs

`libs/code-review/infrastructure/adapters/services/documentation-search-exa.service.ts`:

**(a) Exa client never constructed** — `configService.get('API_EXA_KEY')` empty in the worker process:
```ts
constructor(...) { const apiKey = this.configService.get<string>('API_EXA_KEY'); this.exaClient = apiKey ? new Exa(apiKey) : null; }
async searchByFilePlan(...) { if (!this.exaClient) { this.logger.warn({ message: 'API_EXA_KEY is not configured, skipping documentation search stage' }); return {}; } ... }
```
**(b) Client constructed, but every `exaClient.search(...)` throws** (invalid/rotated key → 401, egress blocked from the worker, `exaCallLimiter`/cache layer failing) and is swallowed:
```ts
} catch (error) { this.logger.warn({ message: `Exa search failed for query: ${task.query}`, ..., error }); return null; }
```
Either way `agent-tools.factory.ts` (~L1491) turns the empty result into `No documentation found for "<pkg>"...`. Note a *successful* search always yields an item (even with 0 results it returns `url: 'unknown'` + a fallback snippet) — and no such item appears anywhere in 8 weeks, so "Exa searched and found nothing" is ruled out; it's (a) or (b).

**To pin it down, grep the production *worker* logs for either string:**
- `API_EXA_KEY is not configured, skipping documentation search stage` → cause (a)
- `Exa search failed for query:` → cause (b) (the attached `error` says why: 401, ENOTFOUND, etc.)

## What is NOT the cause

- Not Exa billing: the account has credit and auto-recharge.
- Not a missing secret at rest: `/prod/kodus-orchestrator/API_EXA_KEY` **exists in AWS SSM** (36-char value). `/qa/...` exists too. (`/homolog/...` does not.)

## Fixes (independent of which cause it is)

1. **Ops**: confirm via the log grep above; then either map `API_EXA_KEY` into the worker env (if (a) — the key is in SSM but may not be declared in the worker task definition; `API_EXA_KEY` is `@optional` in `.env.schema` and appears in no deploy config in this repo), or fix the key/egress (if (b)). Create it for homolog. Re-verify with the Langfuse query above — outputs should start carrying `### <title> **URL:** ...`.
2. **Code**: a disabled/failing tool must not impersonate a negative result. Return distinct, explicit messages for "documentation search is not configured in this environment" and "documentation search errored (<reason>)" vs. a genuine empty result, and tell the agent not to treat the former two as evidence. Or don't register `searchDocs` at all when the client is null (`agent-tools.factory.ts` already gates on `if (documentationSearchService)`). Add a unit test asserting the null-client and the thrown-error paths never produce the `No documentation found` string.

## Impact

Every review since at least late June has run with its only external-verification tool silently non-functional. Expect a measurable share of "third-party API behaves like X" false positives/negatives to trace back to this.

andyst-dev:
Fixes #1762

## Problem
`DocumentationSearchExaService` returns `{}` when no `API_EXA_KEY` is configured (backend never built) and `null` per-query when an Exa call throws (backend broken). Either way the `searchDocs` VERIFY tool rewrote it to **"No documentation found for «pkg» with query «q»"** — semantically *"I searched and there is nothing"*.

The agent could not tell **"tool unavailable/broken"** from **"the library has no docs"**, so every third-party-API claim fell back to the model's prior. In prod this produced false positives (e.g. a valid Jest `--testPathPatterns` flag flagged as invalid) — 100/100 sampled `searchDocs` calls had returned "No documentation found" for weeks.

## Fix
- `DocumentationSearchExaService` exposes `isSearchAvailable()` (backend usable ⟺ an API key is configured).
- Wired through the factory's `DocumentationSearchAdapter` (optional — call sites without a search service are unaffected).
- When unavailable, `searchDocs` returns an explicit **"Documentation search is not available right now… do not conclude the package has no documentation"** instead of a fabricated empty result. Backend-unavailable stays distinct from a genuine empty search.

This addresses the deterministic root-cause (a) — "client never constructed" — which the issue's 8-week data shows is the dominant cause.

## Validation
Regression test in `agent-tools.sandbox-errors.spec.ts`:
- **RED without the fix** — mock backend returns `{}` + claims unavailable → tool reports `No documentation found for "jest"...` (the exact bug).
- **GREEN with the fix** — tool reports `Documentation search is not available right now`.
- 11/11 pass across the sandbox-errors + service specs.

### DIFF
--- a/libs/code-review/infrastructure/adapters/services/documentation-search-exa.service.ts
+++ b/libs/code-review/infrastructure/adapters/services/documentation-search-exa.service.ts
@@ -78,6 +78,15 @@ export class DocumentationSearchExaService {
         this.exaClient = apiKey ? new Exa(apiKey) : null;
     }
 
+    /**
+     * The search backend is usable only when an Exa API key is configured.
+     * Exposed so the agent-facing tool can say "search is unavailable" instead
+     * of "no documentation found" when it never ran (#1762).
+     */
+    isSearchAvailable(): boolean {
+        return this.exaClient !== null;
+    }
+
     async searchByFilePlan(
         planByFile: Record<string, DocumentationQueryPlanByFile>,
         options?: {
--- a/libs/code-review/infrastructure/agents/engine/agent-tools.factory.ts
+++ b/libs/code-review/infrastructure/agents/engine/agent-tools.factory.ts
@@ -1509,7 +1509,13 @@ fi
 
                     const docs = results['agent'] || [];
                     if (docs.length === 0) {
-                        return `No documentation found for "${packageName}" with query "${query}".`;
+                        // A successful Exa search always returns a doc item
+                        // (even a 0-result query yields url:"unknown" + a
+                        // fallback snippet), so empty docs with a configured
+                        // backend means every query failed at runtime (#1762
+                        // cause b) — NOT "the library has no docs". Say so,
+                        // matching the unconfigured-backend branch above.
+                        return `Documentation search ran but returned no documentation for "${packageName}" with query "${query}" (the search backend answered nothing). Do NOT conclude the package has no documentation: this usually means the backend is failing at runtime. If the finding depends on framework documentation, state that verification is unavailable.`;
                     }
 
                     const formatted = docs
--- a/libs/code-review/infrastructure/agents/engine/agent-tools.factory.ts
+++ b/libs/code-review/infrastructure/agents/engine/agent-tools.factory.ts
@@ -47,6 +47,13 @@ export interface DocumentationSearchAdapter {
             }>
         >
     >;
+    /**
+     * True when the search backend is actually usable (e.g. an Exa API key is
+     * configured). Lets the agent-visible tool distinguish "no docs exist"
+     * from "search is not available right now" instead of silently reporting
+     * the former for the latter (#1762).
+     */
+    isSearchAvailable?: () => boolean;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


andyst-dev:
Addressed in commit 8d70dc6. A successful Exa search always returns a doc item, so empty docs with a configured backend means every query failed at runtime (#1762 cause b) — the tool now says 'search ran but returned no documentation; do not conclude the package has no documentation', matching the unconfigured-backend branch. Regression test added (RED with old 'No documentation found', GREEN with the new message); 8/8 sandbox + service specs pass.

PR #1820, opened 2026-09-01
fix(e2e): tolerate unreleased routes in the scheduled self-hosted RBAC matrix

### LINKED ISSUE #1696: Scheduled matrix goes red on unreleased routes: RBAC manifest is from main, image is from the last release
### What happens

The scheduled self-hosted matrix tests the **newest published release** image, with a harness checked out from **main**. That is deliberate and correct — the question a scheduled self-hosted run should answer is "does what customers actually run still work?", and the workflow says so.

But `rbac-authorization` asserts live endpoints against `apps/api/src/controllers/__tests__/rbac-matrix.manifest.json`, which is versioned in the repo and therefore always at main's state. So between the moment someone adds an endpoint and the moment a release ships, the manifest lists routes the tested image does not have, and the scenario fails with `expected 403, got 404`.

Observed on [run 31526446836](https://github.com/kodustech/kodus-ai/actions/runs/31526446836):

```
billing_manager should be DENIED on POST /license/prune-seats (expected 403, got 404)
repo_admin should be DENIED on POST /license/prune-seats (expected 403, got 404)
... 5 mismatches, all 404
```

`POST /license/prune-seats` exists in main and in the manifest. It is not in the last release. Nothing is broken.

### Why it matters

This is fake red, and it recurs on a schedule. Every release cycle that adds a route puts the nightly matrix into a state where the failure is not a defect — and a matrix that cries wolf on a cadence teaches everyone to stop reading it, which is the exact failure mode the current matrix work is trying to end.

It is also invisible in the obvious place: the mismatch is between a checked-in fixture and a container image, so nothing in the diff of either one looks wrong.

### Options

**A. Make the scenario version-aware.** Read the target's build version and treat a route absent from that build as *not verified* rather than failed. Keeps the coverage the scheduled run is there to provide, and keeps a genuine 404 on a route the build DOES have as a real failure. Costs a reliable way to ask the target what it is running.

**B. Restrict the scenario to runs that build the image from the code under test.** Simple, no new machinery, and honest — but the scheduled run loses RBAC coverage against the release, which is precisely the configuration customers are on.

A is better if the target can report its version; B is the fallback if it cannot.

### Note

Whatever is chosen, the same trap applies to any scenario that asserts against a checked-in fixture describing the product's surface. RBAC is the one that has a manifest today; it may not stay the only one.

andyst-dev:
Fixes #1696

## Problem
The scheduled self-hosted matrix tests the **newest published release** image, with the harness checked out from **main** — deliberately, to answer "does what customers actually run still work?". But `rbac-authorization` asserts the committed `rbac-matrix.manifest.json`, which always reflects main. Between an endpoint landing on main and its release, the manifest lists routes the tested image doesn't have, and the cron fails with `expected 403, got 404` — a false red on a healthy release train (observed run 31526446836, 5 mismatches all `POST /license/prune-seats`).

## Fix
`tests/e2e/scenarios/rbac-authorization.ts`:
- When `RBAC_ALLOW_MISSING_ROUTE=1`, a **404** on a manifest route (route not yet in the tested image) is counted as a **release gap**, logged explicitly, and not treated as a verdict mismatch — both in the mutation-deny loop and the GET loop (owner canary 404 short-circuits the whole entry).
- Safety guard: if **over half** the manifest 404s, the run still fails — that's a misprovisioned image (wrong tag), not a release gap, and the cron must not report green against a broken target.

`.github/workflows/e2e-self-hosted-matrix.yml`:
- `RBAC_ALLOW_MISSING_ROUTE` is set only on the **scheduled trigger** (`github.event_name == 'schedule'`). PR and release-train runs test an image built from the same commit as the manifest — there a 404 is a real regression and keeps failing.

## Validation
- TypeScript: `tsc -p tests/e2e/tsconfig.json --noEmit` → 0 errors.
- Workflow YAML parses cleanly; the env diff is purely additive (no existing key touched).
- The change is a live e2e scenario (provisioned matrix), so it is validated by typecheck + logic review here and will be exercised by the scheduled run itself; the half-404 safety guard keeps the tolerance from masking a broken image.

### DIFF
--- a/.github/workflows/e2e-self-hosted-matrix.yml
+++ b/.github/workflows/e2e-self-hosted-matrix.yml
@@ -295,6 +295,12 @@ jobs:
                   IMAGE_TAG: ${{ needs.plan.outputs.image_tag }}
                   SOURCE_MATRIX_FILE: ${{ inputs.matrix_file || 'matrix/fast.yml' }}
                   LICENSE_MODE: ${{ matrix.cell.license }}
+                  # Scheduled cron runs test the newest PUBLISHED release against
+                  # a main-refreshed RBAC manifest, so endpoints added since that
+                  # release legitimately 404. PR / release-train runs test an
+                  # image built from the same commit as the manifest — there a
+                  # 404 is a real regression and must keep failing (#1696).
+                  RBAC_ALLOW_MISSING_ROUTE: ${{ github.event_name == 'schedule' && '1' || '0' }}
                   KODUS_INSTALLER_PATH: ${{ github.workspace }}/kodus-installer
                   # AWS by default: Kodus has credit there, and DigitalOcean
                   # was the direct cause of two failure classes — 504s on
--- a/tests/e2e/scenarios/rbac-authorization.ts
+++ b/tests/e2e/scenarios/rbac-authorization.ts
@@ -98,6 +98,22 @@ async function hit(
     return res.status;
 }
 
+// Guards run before handlers. With a definitely-invalid bearer token the
+// JwtAuthGuard answers 401 when a route EXISTS (it is found and rejects the
+// token) and the router answers 404 when it is ABSENT. Either way no handler
+// runs, so this probe has zero side effects — it is the mutation analogue of
+// the GET branch's owner canary, used to attribute a deny-role 404 to a
+// release gap only when the route is genuinely missing (#1696, rbac:146).
+const UNAUTHENTICATED_PROBE_TOKEN = "orca-rbac-probe-invalid-token";
+
+async function routeExists(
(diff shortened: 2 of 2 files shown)


andyst-dev:
Both findings addressed in commit 5959201. (rbac:146) A deny-role 404 is now confirmed against route existence first — an unauth probe answers 401 if the route exists (guard runs, no side effect) vs 404 if absent — so an existing route that 404s stays a real failure; only a genuinely absent route is tolerated. (rbac:170) The safety guard now counts distinct missing endpoints in a Set, not per-role 404s, so the ratio can't be inflated.

PR #1835, opened 2026-09-02
fix(log): surface statusCode/responseBody on BYOK errors

### LINKED ISSUE #1829: BYOK provider errors: statusCode and responseBody dropped from logs, user sees 'Unexpected error'
## Summary

When a BYOK provider call fails (400/401/404), our logs keep only `name`/`message`/`stack` of the `AI_APICallError`. The provider's `statusCode` and `responseBody` — the only actionable part — are dropped, so the customer sees "Unexpected error while running the code review (openai_compatible)" and we cannot tell from logs whether the failure was the gateway, the model id, or us.

## Where it is lost

- **Agent path:** the runner does capture `status` + `responseBody` into the trace event (`libs/agent-harness/infrastructure/ai-sdk/ai-sdk-agent-runner.ts:312-323`) and `provider-error.ts` re-attaches them to the reconstructed Error. But the log emitted in `AgentReviewStage` / orchestrator (`[AGENT] <agent> failed for PR#N after Xms: Bad Request`) serializes the error object without those properties — production log attributes for this message are only `prNumber, durationMs, model, errorName, errorStack`.
- **Summary path:** `generateOverallComment` logs the raw `AI_APICallError` (`Error generateOverallComment pull request: PR#N`) and the serializer again keeps only name/message/stack.
- Langfuse observations for these calls carry only `statusMessage: "Bad Request"`.

## Repro

Point an `openai_compatible` slot at any endpoint that returns a 400 with a JSON body (e.g. an invalid model id on a gateway) and run a review. Check the logs: no `responseBody`, no `statusCode`.

## Proposed fix

- In both log sites, log `statusCode`, `responseBody` (truncated to ~2KB) and the request `url` host as explicit metadata fields, not inside the serialized Error.
- Make the error serializer (`libs/core/log`) include enumerable own props of Error subclasses (`statusCode`, `responseBody`, `url`) generically, so other call sites benefit.
- Propagate the classified reason into the PR "review failed" reaction/comment (this was the intent of #1568; today the fallback text still reaches the user).

## Why it matters

We spent a day on a churned-customer investigation and could not determine from our own logs whether two 400s came from the customer's gateway config or from our request shape. This is a half-day fix that turns every future BYOK support ticket into a one-query answer.

andyst-dev:
## Summary

BYOK provider errors attach `statusCode`, `responseBody` and `url` as own props of the `Error` subclass, but the shared log serializer only kept `name/message/stack` (prototype fields) — so every structured log line dropped the only actionable part and the user was left with "Unexpected error while running the code review". This wires those fields through `SimpleLogger`'s serializer generically, so **every** log site (agent-review stage, `generateOverallComment`, …) now surfaces them with no per-call changes.

## What & why

`Fixes #1829` — see the issue for the full investigation. The runner already captures `status`/`responseBody` into the trace event, but:

- **Agent path:** `AgentReviewStage` / orchestrator logs `[AGENT] <agent> failed for PR#N after Xms` and serializes the error as `{ name, message, stack }` — the provider's `statusCode`/`responseBody` never reach the log.
- **Summary path:** `generateOverallComment` logs the raw `AI_APICallError`, same serializer loss.

The fix:

1. **`extractErrorProps(error, maxLen)`** — new helper in `libs/core/log/logger.ts`. Merges enumerable own props of an `Error` subclass (excludes `name`/`message`/`stack`, which live on the prototype), redacts values through `deepSanitize`, and truncates string values to ~2 KB.
2. **`SimpleLogger.buildLogObject`** — now spreads `extractErrorProps` into the `error` object it emits. Since this is the shared serializer, all call sites benefit at once.

No behavior change when the error has no extra own props (returns `{}`).

## Test plan

- New unit spec `libs/core/log/extract-error-props.spec.ts` (3 cases):
  - extracts `statusCode`/`responseBody`/`url` from a BYOK-style `Error` subclass, excludes prototype fields
  - truncates an oversized `responseBody` to ~2 KB + ellipsis, leaves small values intact
  - returns `{}` for a plain `Error` with no extra own props
- `jest libs/core/log/extract-error-props.spec.ts --runInBand` → **3 passed**
- `pnpm tsc --noEmit` → clean (0 errors across the monorepo)
- `eslint` on the two changed files → clean

Note: this covers the "serializer drops the fields" half of #1829. Propagating the classified reason into the per-PR review-failure comment is the follow-up flagged in the issue.


Wellington01:
@andyst-dev 

This grabs every own enumerable prop on the Error, not just statusCode/responseBody/url. The actual class this targets, @ai-sdk/provider's APICallError, also carries `requestBodyValues` (the full request body — for a review call that's the whole prompt, messages array included) and `data`. Both are objects, so they skip the string-length branch entirely and go straight to `deepSanitize(value)` with no size cap of any kind. Every provider error (429, 500, timeout) is now going to dump the full LLM request payload into the structured log, uncapped.

Scope this to an allowlist (statusCode, responseBody, url, maybe responseHeaders) instead of Object.keys(error) minus name/message/stack.

libs/core/log/logger.ts:708-711

Sanitize order bug: values under maxStringLength go through deepSanitize (line 711), values over it just get substring()'d (line 709) with no sanitize at all. Any credential-shaped string that happens to be long skips redaction entirely. Should sanitize first, then truncate the result.

extract-error-props.spec.ts only exercises a 3-field toy subclass. Worth adding a case against the real APICallError shape (or at least one with an extra `requestBodyValues`-like object prop) so the allowlist/cap regresses if it's ever loosened.

andyst-dev:
Thanks @Wellington01 — all three points addressed in `747433e`:

1. **Allowlist instead of `Object.keys(error)`** — `extractErrorProps` is now scoped to a `Set` of `statusCode`, `responseBody`, `url`, `responseHeaders`. `APICallError`'s `requestBodyValues` (the full request payload) and `data` no longer reach the log at all, so the uncapped object dump is gone. Added a regression test with an `APICallError`-shaped subclass carrying `requestBodyValues`/`data` as large object props, asserting they're dropped.

2. **Sanitize order** — strings now go through `sanitizeString` (URL credential redaction) **before** the `maxStringLength` truncation, so an oversized credential-shaped string can't skirt redaction by being truncated first. Regression test: a `mongodb://user:sekret@…` URL padded past the cap must not contain `sekret` and must carry `[REDACTED]`.

3. **Test coverage** — added both cases above to `extract-error-props.spec.ts` (real `APICallError`-like shape with heavy object props, plus the sanitize-before-truncate case) alongside the existing toy-subclass coverage.

Validation:
- `API_NODE_ENV=test pnpm exec jest libs/core/log/extract-error-props.spec.ts --config jest.config.ts --forceExit --runInBand` → **5/5 pass**. TDD: the two new tests fail against the previous `Object.keys` implementation (2 failed), green with the allowlist.
- `eslint` and `tsc --noEmit` on both touched files: clean.
- Head `747433e` pushed (`mergeable` recomputing).

andyst-dev:
Good catch @kody — the allowlist was too narrow. Extended `ERROR_LOG_PROPS` in `b66a65c` with the in-repo scalar diagnostics operators relied on, and updated the still-generic `buildLogObject` comment.

Added props (all scalar number/string, so they pass through the sanitize/deepSanitize/truncate path safely and stay small):
- `status` — azure-repos-request-helper.ts:603
- `gate` / `target` / `requestId` — code-review-job-processor.service.ts:248/254
- `code` — mcp-protocol.utils
- `contextWindow` / `overheadTokens` / `estimatedTokens` / `contextWindowTokens` / `modelName` — libs/llm/errors.ts (AgentContextWindowTooSmallError / AgentPromptTooLargeError)

`requestBodyValues`/`data` remain excluded (the payload dump is the whole point of the allowlist).

Validation:
- New regression test: an in-repo-style error carrying `status`/`contextWindow`/`modelName` plus a heavy `requestBodyValues` → scalars surfaced, payload dropped. **RED** against the narrowed allowlist, **GREEN** with the extension.
- `extract-error-props.spec.ts` → **6/6 pass**; eslint + prettier clean on both files.
- Head `b66a65c` pushed.

andyst-dev:
Fair, and it is the stronger half of the note: the cap only existed in the string branch, so a non-scalar value walked straight past it. Fixed in `2c89695` by taking your second option rather than dropping `target` from the allowlist, since the size bound is the invariant worth having for every entry, not just this one.

`extractErrorProps` now caps the serialized form of non-string values too: a value that already fits keeps its original shape (numbers stay numbers, a compact `target` stays an object and remains queryable in the log pipeline), and an oversized one degrades to a truncated JSON string. `JSON.stringify` is wrapped in a try/catch so a circular or otherwise non-serializable value is returned untouched instead of dropped.

I also corrected the allowlist comment, which claimed every entry was scalar (`number`/`string`) while `target` is the `CommandReviewFeedbackTarget` object.

Validation: `extract-error-props.spec.ts` is 8/8, with two new cases (an oversized `target` capped to 2 000 + ellipsis, a compact one left intact and `statusCode` still a number). RED verified on the pre-fix extractor: 7 passed / 1 failed, only the oversized case failing. eslint clean and prettier clean on both files.

Wellington01 (approved):

PR #1898, opened 2026-09-10
fix(llm): classify 5xx phrasing as TRANSIENT for BYOK fallback (#1875)

### LINKED ISSUE #1875: Agent-based review skips BYOK fallback: reconstructed agent error loses HTTP status and classifies as UNKNOWN (2.2.1)
## Summary

In `selfhosted-2.2.1`, the runtime model failover (`primary → fallback`) added for the agent-based review flow (PR #1479 / issue #1430) is **defeated by error misclassification** when the failure comes from the agent path. The configured BYOK fallback is never tried for "Agent-Based Code Review" / "Generalist Agent" steps, because the error that reaches `shouldFailoverToNextModel` has lost its HTTP status and doesn't match any message-based rule — so it classifies as `UNKNOWN`, and `UNKNOWN` deliberately never cascades.

The same org, same outage, same fallback config: the **summary step works** (cascades to fallback), the **agent steps don't**. That asymmetry is the tell.

## Environment

- Self-hosted, image `ghcr.io/kodustech/kodus-ai-*:2.2.1`
- BYOK `openai_compatible` provider, `main` + `fallback` configured at org level
- Failure mode: main provider behind a proxy returning `502 Bad Gateway` / Cloudflare-style `530`

## What we observed (worker logs, 2.2.1)

Generalist agent — fallback **skipped**, classified `UNKNOWN`:

```
[LLM-ERROR] code-review-generalist: "smart" failed (UNKNOWN) — Failed after 4 attempts. Last error: AI_APICallError: Bad Gateway
```

Summary step from the **same run, same error** — fallback correctly used:

```
[LLM-ERROR] [model-failover] generateSummaryPR: "smart" failed (TRANSIENT); cascading to fallback "accounts/fireworks/models/..."
```

The agent then surfaces the `UNKNOWN`-category friendly message to the user:

> Unexpected error while running the code review (openai_compatible).

…which is literally the `buildFriendlyMessage` default for `UNKNOWN`.

## Root cause

Two pieces interact in `selfhosted-2.2.1`:

1. **`libs/code-review/infrastructure/agents/collaborators/provider-error.ts`** — the agent harness swallows thrown provider errors into a result (`finishReason: 'error'`), and `providerErrorFromResult` reconstructs a throwable. It re-attaches `statusCode`/`responseBody` **only if the harness captured them** (`result.errorStatus`). For a retry-exhausted failure the harness result carries only `errorMessage`/`errorName`, so the reconstructed error is effectively:

   ```
   AI_RetryError: Failed after 4 attempts. Last error: AI_APICallError: Bad Gateway
   ```

   …with **no `status`/`statusCode`/`cause`** on it.

2. **`libs/llm/error-classifier.ts`** — with no HTTP status to read (`extractHttpStatus` → `undefined`), classification falls to `matchByMessage`. The `TRANSIENT` token list knows `econnreset`, `etimedout`, `socket hang up`, `network error`, `fetch failed`, `timeout`, `aborted` — but **not** `bad gateway`, `502`, `503`, `service unavailable`, or `530`. The reconstructed text matches nothing → `UNKNOWN`.

3. `shouldFailoverToNextModel` then correctly-per-design returns `false` for `UNKNOWN` ("stay conservative rather than spend a 2nd billed call") — so the org's fallback never runs, and a pure infra outage on the main provider fails the whole agent review even though a healthy fallback was configured.

Note the summary path doesn't hit this because it throws the raw Vercel AI SDK `AI_APICallError`, which carries `status: 502` → `matchByHttpStatus` → `TRANSIENT` → cascade.

## Suggested fixes (either, or both)

1. **Preserve the status through the harness boundary**: when the agent-loop result is built from a caught error, capture `errorStatus`/`errorResponseBody` from the original error (including for `AI_RetryError`, e.g. from the last attempt or the `cause` chain), so `providerErrorFromResult` can re-attach it and `matchByHttpStatus` can do its job.
2. **Widen the `TRANSIENT` message vocabulary** in `matchByMessage` to include `bad gateway`, `502`, `503`, `service unavailable`, `bad gateway`-proxied Cloudflare `530`, etc. Status-code matching is the reliable half of the classifier; the message list should be a safety net that covers the common 5xx phrasings too.

Alternatively, an `AI_RetryError` whose message embeds a `... Last error: AI_APICallError: <status text>` could be re-classified from that embedded last error.

## Impact

- Orgs with a healthy BYOK fallback still get **failed reviews + ❌ reaction on the PR** whenever the main provider has a transient 5xx outage lasting longer than the SDK's 4 same-model retries.
- Misleading UX: user sees the generic "Unexpected error while running the code review" instead of a transient-provider message, and may wrongly suspect their fallback config.

Related: #1430 (closed by #1479 — the failover plumbing works; this is the classification layer defeating it).

andyst-dev:
Fixes #1875

## Problem
The agent-based review path defeats the runtime model failover (`primary → fallback`) on a transient upstream outage. When the main provider dies with a 5xx, the AI SDK wraps retry exhaustion in an `AI_RetryError` whose default shape is a **plain `Error`** — `message: 'Failed after 4 attempts. Last error: AI_APICallError: Bad Gateway'`, with **no `status`/`statusCode`/`cause`**. So `classifyLLMError` reads no HTTP status, and `matchByMessage` — which knew `econnreset`/`etimedout`/`socket hang up`/`fetch failed`/`timeout`/`aborted` but **not** `bad gateway`/`service unavailable` → labelled it `UNKNOWN`. `shouldFailoverToNextModel` deliberately never cascades `UNKNOWN`, so the org's configured BYOK fallback is never tried and a transient infra outage fails the whole agent review with a misleading `❌`.

The asymmetry is the tell: the **summary** step cascades (`matchByHttpStatus` sees status 502 on the raw `APICallError` it throws), the **agent** steps don't (status stripped by the RetryError wrap).

## Fix
Extend the **TRANSIENT message vocabulary** in `matchByMessage` with the common 5xx phrasings — `bad gateway`, `gateway timeout`, `service unavailable`, `internal server error`, plus the status texts `502`/`503`/`504`/`530` (Cloudflare's code-over-5xx). This mirrors the reliable `matchByHttpStatus` half (`status >= 500` → TRANSIENT) for the message-only fallback path, so a status-less RetryError-embedded 5xx now classifies TRANSIENT → `shouldFailoverToNextModel` cascades → the BYOK fallback runs. Redundant for errors that DO carry a status (those were already handled); it only closes the message-only gap.

## Testing
- `error-classifier.spec.ts` — new `it.each` over status-less 5xx phrasings (incl. the exact `'Failed after 4 attempts. Last error: AI_APICallError: Bad Gateway'` from the report) → **TRANSIENT**. TDD verified: **RED** on the old vocabulary (5 failed), **GREEN** with the fix — **60/60** suite pass, `eslint` clean.

### DIFF
--- a/libs/llm/error-classifier.spec.ts
+++ b/libs/llm/error-classifier.spec.ts
@@ -226,6 +226,24 @@ describe('classifyLLMError', () => {
             );
         });
 
+        it.each([
+            // An AI_RetryError (SDK retries exhausted) embeds the last attempt
+            // as text with NO status field — the default RetryError is a plain
+            // Error. These must still classify as transient so the configured
+            // BYOK fallback cascades instead of failing the whole review (#1875).
+            'Failed after 4 attempts. Last error: AI_APICallError: Bad Gateway',
+            'Last error: AI_APICallError: 502 Bad Gateway',
+            'Upstream error: service unavailable',
+            'Proxy returned 530 (Bad Gateway) from Cloudflare',
+            'An upstream error occurred: internal server error',
+            'gateway timeout while waiting for the model to respond',
+        ])('5xx phrasing without a status (%s) → TRANSIENT', (msg) => {
+            // No `status`/`statusCode` on the error — message-string fallback only.
+            const err = new Error(msg);
+            const info = classifyLLMError(err);
+            expect(info.category).toBe(LlmErrorCategory.TRANSIENT);
+        });
+
         it('unrecognized message → UNKNOWN', () => {
             const err = new Error('something weird happened in the SDK');
             expect(classifyLLMError(err).category).toBe(
--- a/libs/llm/error-classifier.spec.ts
+++ b/libs/llm/error-classifier.spec.ts
@@ -244,6 +244,19 @@ describe('classifyLLMError', () => {
             expect(info.category).toBe(LlmErrorCategory.TRANSIENT);
         });
 
+        it.each([
+            // A 5xx-like number embedded in a larger number is not a status:
+            // substring checks used to read "5032" as a 503 and flip a permanent
+            // failure to TRANSIENT, defeating the purpose of the fallback.
+            'The input token count (5032) exceeds the maximum allowed for this model',
+            'Request id req_5301 rejected by the gateway',
+        ])('bare digit inside a larger number (%s) → not TRANSIENT', (msg) => {
+            const err = new Error(msg);
+            expect(classifyLLMError(err).category).not.toBe(
+                LlmErrorCategory.TRANSIENT,
+            );
+        });
+
         it('unrecognized message → UNKNOWN', () => {
             const err = new Error('something weird happened in the SDK');
             expect(classifyLLMError(err).category).toBe(
--- a/libs/llm/error-classifier.ts
+++ b/libs/llm/error-classifier.ts
@@ -394,11 +394,10 @@ function matchByMessage(lower: string): LlmErrorCategory {
         lower.includes('service unavailable') ||
         lower.includes('internal server error') ||
         // Status numbers can appear as text without a status field (Cloudflare's
-        // 530 over a 5xx, proxy passthrough).
-        lower.includes('502') ||
-        lower.includes('503') ||
-        lower.includes('504') ||
-        lower.includes('530')
+        // 530 over a 5xx, proxy passthrough). Word boundaries keep bare digits
+        // inside larger numbers (token counts like "5032", ids) from being read
+        // as an HTTP status and wrongly marking a permanent error TRANSIENT.
+        /\b(?:502|503|504|530)\b/.test(lower)
     ) {
         return LlmErrorCategory.TRANSIENT;
     }
--- a/libs/llm/error-classifier.ts
+++ b/libs/llm/error-classifier.ts
@@ -382,7 +382,23 @@ function matchByMessage(lower: string): LlmErrorCategory {
         lower.includes('network error') ||
         lower.includes('fetch failed') ||
         lower.includes('timeout') ||
-        lower.includes('aborted')
+        lower.includes('aborted') ||
+        // Common 5xx phrasings. `matchByMessage` is the fallback when no HTTP
+        // status reached the classifier — e.g. an AI_RetryError (SDK retries
+        // exhausted) whose embedded text "Last error: AI_APICallError: Bad
+        // Gateway" carries no status (the default RetryError is a plain Error).
+        // Without these, a real upstream outage classifies UNKNOWN and the
+        // configured BYOK fallback is never tried (#1875).
+        lower.includes('bad gateway') ||
+        lower.includes('gateway timeout') ||
+        lower.includes('service unavailable') ||
+        lower.includes('internal server error') ||
+        // Status numbers can appear as text without a status field (Cloudflare's
+        // 530 over a 5xx, proxy passthrough).
+        lower.includes('502') ||
+        lower.includes('503') ||
+        lower.includes('504') ||
+        lower.includes('530')
     ) {
         return LlmErrorCategory.TRANSIENT;
     }


andyst-dev:
Good catch. Swapped the four bare substring checks for a word-boundary regex, so a digit sequence inside a larger number (like the token count 5032) no longer reads as an HTTP status. Added two regression cases for it. All 62 classifier tests pass.

andyst-dev:
You are right, and it is the worse half of the two bugs: `\b` is defined over `[A-Za-z0-9_]`, so the status glued to a letter or underscore stopped matching, those messages classified UNKNOWN, and `shouldFailoverToNextModel` deliberately returns false for UNKNOWN. Net effect: the BYOK fallback stayed switched off for the exact transient outage this branch was added for. Fixed in `60705b0`.

I took your lookbehind form: `/(?<!\d)(?:502|503|504|530)(?!\d)/`. It keeps the original goal (a digit sequence inside a larger number is not a status: `5032`, `req_5301`) while matching `HTTP_503`, `ERR_502`, `upstream_530` and `http503`. Lookbehind is fine here: the project targets ES2022 and the test runtime is Node 26, so no need for the `(?:^|\D)` alternative.

Regression cases added for the glued forms (`HTTP_503`, `ERR_502`, `upstream_530`, `http504`, `502badgateway`), all of which classified UNKNOWN before. RED verified on the previous regex (62 passed / 5 failed, only the new cases failing), GREEN now: `error-classifier.spec.ts` 67/67. eslint clean on both files.

malinosqui (changes requested):
Thanks for this! The new phrases (`bad gateway`, `service unavailable`, ...) are the right safety net and fix the exact message from #1875. My concern is the numeric regex: it also matches status codes inside ids and model names, so some permanent errors would be classified as `TRANSIENT`.

### False positives in the numeric regex

`matchByMessage` doesn't only see status-less 5xx errors. `matchByHttpStatus` returns `UNKNOWN` for any status it doesn't map, like 400 or 422, and `classifyLLMError` then falls back to `matchByMessage` using the full error text, response body included ([error-classifier.ts#L113-L116](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/llm/error-classifier.ts#L113-L116)). With this PR, a permanent 400 whose body happens to contain one of these strings becomes `TRANSIENT`:

| input (lowercased error text) | this PR |
|---|---|
| `invalid parameter temperature (request id 8f3c1a2b-503a-4c1d-9e2f-0a1b2c3d4e5f)` | TRANSIENT |
| `request_id: req_503ab91` | TRANSIENT |
| `model qwen3-530b is not supported` | TRANSIENT |
| `trace-504x9 invalid schema` | TRANSIENT |
| `id 503-abc` | TRANSIENT |

`(?<![a-z0-9])` still allows `-` and `_` before the number, and `(?!\d)` allows letters after it. `shouldFailoverToNextModel` always cascades on `TRANSIENT` ([model-failover.ts#L84](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/llm/model-failover.ts#L84)), so each case above spends a second billed call on the fallback. That call will most likely fail the same way, and the user sees "provider temporarily unavailable" instead of the real error.

One possible tightening: no word character or `-` on either side of a bare number, with an explicit keyword required for the glued forms:

```ts
/(?<![\w-])(?:502|503|504|530)(?![\w-])|(?<![a-z0-9])(?:http|err|error|status|code)[_-]?(?:502|503|504|530)(?![\w-])/
```

I ran it against the cases below. It keeps `upstream returned 503`, `HTTP_502`, `http-504`, `status: 530`, `(503) service`, `error 503.`, `ERR_502` and `code_503`, and rejects every false positive in the table. Dropping the numbers entirely and relying on the phrases would also work for #1875.

### Requested changes

1. Tighten the numeric regex as above (or remove it), and add negative cases to `error-classifier.spec.ts`: UUID segment, `req_503…`, a model name like `qwen3-530b`, `503-abc`. Today the spec only asserts positives, so nothing catches this.

### Suggested follow-up (fine as a separate PR)

The root cause in #1875 is that the status never reaches the classifier. `AI_RetryError` keeps the real `APICallError` in `lastError` (and `errors[]`), but the agent runner only reads `statusCode` and `status` on the top-level error ([ai-sdk-agent-runner.ts#L315-L320](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/agent-harness/infrastructure/ai-sdk/ai-sdk-agent-runner.ts#L315-L320)). If it also fell back to `lastError.statusCode` and `lastError.responseBody`, `matchByHttpStatus` would classify these as 5xx from the real status, with no text heuristics involved. The message vocabulary would then only be a safety net.

PR #1901, opened 2026-09-10
fix(platformData): cap embedded diffs below MongoDB's 16MB BSON ceiling (#1841)

### LINKED ISSUE #1841: Very large PRs fail code review: single pullRequests document exceeds MongoDB's 16 MB limit
## Summary

Code review of a **very large pull request** fails because Kodus persists each PR as a **single MongoDB document** that embeds every file's diff/patch plus all commits. On a large PR the aggregate document exceeds MongoDB's hard **16 MB (16,777,216‑byte) BSON document limit**, so the write is rejected and the review cannot complete.

The user‑visible result is an inconclusive review — either the check reports **"Code Review Skipped — Unable to load or resolve the review configuration"**, or it appears to hang, depending on where in the pipeline the failed write lands.

## Environment

- Deployment: self‑hosted via Docker Compose (bundled datastores)
- Kodus images: `2.1.33` (api / worker / webhook / web)
- Orchestrator: `2.1.8`
- MongoDB: `mongo:8`
- RabbitMQ: `4.2.2-kodus`

## Error

The worker throws while writing the PR document:

```
MongoBulkWriteError: Plan executor error during update ::
  caused by :: Resulting document after update is larger than 16777216
```

Observed the offending write directly in the MongoDB slow‑query log: a single `update` on the `pullRequests` collection whose response length was `16777236` — i.e. ~20 bytes over the 16 MB ceiling. The update batches ~200 sub‑updates, each setting `files.$.patch` to a file's full unified diff, into one document.

## Root cause

`PullRequestsService` (`bulkApplyFileChanges`) stores the PR and **all** of its file patches in one `pullRequests` document. For a large PR — e.g. a release/promotion PR that carries the whole accumulated diff across hundreds of files and commits — the embedded patches push the document past MongoDB's per‑document 16 MB limit. This is a structural size ceiling in MongoDB, so it cannot be worked around with more memory or a longer timeout.

Because the persist fails, downstream stages that read the PR/config error out. Representative stack (Kodus source frames):

```
PullRequestsService.bulkApplyFileChanges         → MongoBulkWriteError (doc > 16 MB)
...
ResolveConfigStage.executeStage
  → CodeBaseConfigService.getConfig
    → KodyRulesRepository.findByOrganizationId    → throws
→ review returns "Skipped — Unable to load or resolve the review configuration"
```

## Secondary robustness issue (related)

A code‑review job that cannot complete within RabbitMQ's default consumer acknowledgement timeout (30 min) causes the broker to close the channel:

```
Channel closed by server: 406 (PRECONDITION-FAILED) with message
"PRECONDITION_FAILED - delivery acknowledgement on channel N timed out.
 Timeout value used: 1800000 ms."
```

RabbitMQ then **requeues** the message, the worker re‑consumes it, re‑runs the entire (still‑failing / still‑too‑long) review, and this repeats **indefinitely** — the same PR is reviewed over and over with no dead‑lettering or redelivery cap. A large PR (or any job that legitimately runs >30 min) triggers this loop. Consider:

- a dead‑letter queue / max‑redelivery cap for `handleCodeReviewJob`, and/or
- raising `consumer_timeout` for the review queue and/or ack’ing early with heartbeated progress.

## Suggested fixes for the primary issue

- Store file patches **out of the PR document** — one document per file (e.g. a `pullRequestFiles` collection keyed by PR id), or use **GridFS** for large patches, so the `pullRequests` document stays well under 16 MB.
- Enforce a per‑file patch size cap and/or a total‑diff cap, skipping/truncating oversized files with a visible note in the review, rather than failing the whole review.
- Fail closed with a clear, actionable status (e.g. "PR too large to review — N files / X MB diff") instead of the generic "Unable to load or resolve the review configuration".

## Impact

Large PRs (release/promotion PRs, big refactors) silently get no review. Because the failed write can also spike memory on a resource‑constrained bundled MongoDB, it can additionally OOM the mongo container in small self‑hosted deployments.

andyst-dev:
Fixes #1841

## Problem
A very large PR is persisted as a single `pullRequests` MongoDB document that embeds every file's unified diff. When the aggregate passes MongoDB's hard 16 MB per-document BSON ceiling, the write is rejected (`MongoBulkWriteError ... Resulting document after update is larger than 16777216`) and the whole review fails with a misleading **"Code Review Skipped — Unable to load or resolve the review configuration"**. This is a structural ceiling — it can't be fixed with more memory or a longer timeout.

## Fix — bound the embedded diff, don't fail the review
The document always stays under the cap; oversized diffs degrade (with a visible marker) instead of breaking the run.

- **`diff-budget` util** (new): `MAX_PATCH_CHARS_PER_FILE` (200 000) plus a shared **aggregate** `MAX_TOTAL_EMBEDDED_PATCH_CHARS` (8 MB — headroom below the 16 MB BSON cap for the rest of the doc: suggestions, commits, metadata). `budgetPatchForPersist` truncates an oversized patch to its per-file/aggregate allowance (keeping a leading window **plus a visible marker**) or drops it once the aggregate budget is spent.
- **`PullRequestsService.handleExistingPullRequest`**: seeds a running aggregate budget with the bytes **already embedded** on the doc's existing files (so the whole document, not just this batch, is bounded), then routes every `updateFile` patch through the budget. Capped files are flagged `patchTruncated: true`.
- **Repository `translateFileBulkOp`**: last-resort per-file clamp on any `files.$.patch` written by any path — a future caller that bypasses the service budget still cannot embed an unbounded patch.
- **Types**: `IFile` + `PullRequestsModel.files` now type `patch?` / `patchTruncated?`.

Net: a release/Promotion PR over 16 MB now reviews with diff coverage trimmed at the tail (each affected file carries the marker), instead of failing with no result at all.

## Validation
- `diff-budget.spec.ts` (9): per-file cap, aggregate budget, drop-on-exhaustion, marker presence, cumulative-consumed ≤ budget, repo clamp.
- `pullRequests.service.bulk-file-changes.spec.ts`: 2 new wiring tests — (a) a 2×-cap patch is truncated + marked, (b) an already-embedded doc (budget ~exhausted) truncates the first file and **drops** the next. Adjusted one pre-existing case that asserted an oversized text patch is stored verbatim — the behaviour #1841 intentionally bounds.
- **30/30 pass**; `tsc` clean on all touched files; `eslint` clean. **RED verified**: stashing the service wiring makes both budget tests fail (the cap is not a no-op).

## Scope left / notes
- The secondary **RabbitMQ 30-min ack-timeout re-loop** from the issue (dead-letter/cap + `consumer_timeout`) is a separate concern and out of scope here — should be its own PR.
- "Fail closed with an explicit status" is now moot for document-size (we never fail); still applies to genuinely-unreviewable PRs and is separate.
- Constants are deliberate (no env-tuning), matching the existing `MAX_FILES_PER_SAVE` policy in the same file.

### DIFF
--- a/libs/platformData/domain/pullRequests/contracts/pullRequests.repository.ts
+++ b/libs/platformData/domain/pullRequests/contracts/pullRequests.repository.ts
@@ -236,6 +236,17 @@ export interface IPullRequestsRepository {
         totalChanges: number;
     }>;
 
+    /**
+     * Server-side sum of the UTF-8 bytes embedded in `files[].patch` for one
+     * PR (`$strLenBytes`). Lets the caller enforce the aggregate embedded-diff
+     * budget against ground truth after a write instead of trusting a
+     * potentially stale in-memory read of `existingPR.files` (#1841).
+     */
+    computeEmbeddedPatchBytes(
+        prUuid: string,
+        organizationId: string,
+    ): Promise<number>;
+
     /** Generates a stable id for file/suggestion sub-documents. */
     newSubDocumentId(): string;
 
--- a/libs/platformData/domain/pullRequests/interfaces/pullRequests.interface.ts
+++ b/libs/platformData/domain/pullRequests/interfaces/pullRequests.interface.ts
@@ -193,6 +193,13 @@ export interface IFile {
         generateSuggestions: string;
         safeguard: string;
     };
+    // Unified diff persisted for review context. Written at save-time and
+    // consume a shared aggregate budget so the PR document stays under
+    // MongoDB's 16 MB BSON ceiling; `patchTruncated` marks a diff that was
+    // capped (the patch carries a trailing marker) rather than persisted whole
+    // (#1841).
+    patch?: string;
+    patchTruncated?: boolean;
 }
 
 export interface IPullRequestUser {
--- a/libs/platformData/domain/pullRequests/utils/diff-budget.spec.ts
+++ b/libs/platformData/domain/pullRequests/utils/diff-budget.spec.ts
@@ -1,18 +1,30 @@
 import {
-    MAX_PATCH_CHARS_PER_FILE,
-    MAX_TOTAL_EMBEDDED_PATCH_CHARS,
+    MAX_PATCH_BYTES_PER_FILE,
+    MAX_TOTAL_EMBEDDED_PATCH_BYTES,
     TRUNCATED_DIFF_MARKER,
     budgetPatchForPersist,
     clampPatchForPersist,
+    clampPatchForPersistWithFlag,
+    utf8ByteLength,
 } from './diff-budget';
 
 describe('diff-budget (#1841)', () => {
+    describe('utf8ByteLength', () => {
+        it('counts UTF-8 bytes, not UTF-16 code units', () => {
+            expect(utf8ByteLength('abc')).toBe(3);
+            // CJK: 3 bytes per char in UTF-8, 1 code unit in UTF-16.
+            expect(utf8ByteLength('漢字')).toBe(6);
+            // Emoji outside the BMP: 4 bytes, 2 UTF-16 code units.
+            expect(utf8ByteLength('😀')).toBe(4);
+        });
+    });
+
     describe('budgetPatchForPersist', () => {
         it('passes a patch that fits the per-file cap through unchanged', () => {
             const patch = 'x'.repeat(1000);
             const result = budgetPatchForPersist(
                 patch,
-                MAX_TOTAL_EMBEDDED_PATCH_CHARS,
+                MAX_TOTAL_EMBEDDED_PATCH_BYTES,
             );
             expect(result.patch).toBe(patch);
             expect(result.truncated).toBe(false);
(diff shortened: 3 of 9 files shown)


andyst-dev:
Addressed all four. The budgets are now measured in UTF-8 bytes rather than string length, so multibyte diffs are capped against what MongoDB really stores. The aggregate budget re-credits a file's old patch before charging the replacement, so a re-sync no longer double-counts and wipes the tail files. patchTruncated is always written explicitly, so a file that fits no longer keeps a stale flag. And after the bulk write it re-reads the real embedded size and retries with a tighter budget if it overflowed, instead of trusting the pre-write tally. 39 diff-budget tests pass.

andyst-dev:
Addressed all three from the second round. Pushed in `0d4096ce7`.

1. **Storage clamp escalation could be clobbered.** In `translateFileBulkOp` the `$set['files.$.patchTruncated'] = true` was written inside the `k === 'patch'` branch, so a `patchTruncated: false` entry later in the same payload overwrote it through the generic `$set` line, leaving a capped diff marked as complete. The escalation is now collected in a local flag and applied after the loop, so no later key can revert it.

2. **Failed first-pass patch ops inflated the retry budget.** `ourPatchBytes` still counted bytes from patch ops the bulk write rejected (their `opIndex` is in `bulkResult.errors`). The enforcement pass now maps `errors[].opIndex` back to the patch ops we emitted and subtracts those bytes before computing `tightened`, so the retry gets the real remaining headroom instead of re-writing past the ceiling. The per-op byte map is built at op-construction time.

3. **organizationId missing on the ceiling logs.** Added to the metadata of the three embedded-diff warning logs (skipped verification, "owns none of it", tighter-budget retry) so the warnings correlate to a tenant.

Validation:

- `npx jest libs/platformData` -> 7 suites, 131 passed (was 128; 3 new regression tests).
- Two new tests are RED->GREEN: reverting only the source changes fails the clamp-escalation test and the failed-op budget test, both pass with the fix.
- `tsc -p tsconfig.json` clean on the touched files (the remaining tsc output is pre-existing on main, untouched by this diff).
- PR still reads `mergeable: true`.

andyst-dev:
Good catch, and you're right — the reconciliation only ran against the initial bulkWrite's errors. The retry loop reassigns `ourPatchBytes = consumed` on each pass, so a patch op a retry rejected was still counted as embedded and inflated the next `tightened`, which could burn all three enforcement passes with the doc still over the 16MB ceiling.

Fix in `839d99019`: extracted `reconcileFailedPatchBytes` and call it after **every** write — the initial one and each retry — building the op→patch-byte map from the ops that were just sent (the `updateOps` each carry their patch payload). Now the retry's own failures are credited back before the next iteration.

Validation: `npx jest libs/platformData` → 7 suites, 131 passed (no new test added: the existing "subtracts the bytes of patch ops that FAILED the first bulk write" test exercises the same helper and the refactor doesn't change the first-write path). tsc clean on the touched file. Head confirmed `839d99019`, `mergeable: true`.

Thanks for the careful read.

andyst-dev:
Good catch — and you're right that my round-3 reconciliation was a regression. Applied the correction in `121c331b9`.

A rejected **retry** op does not remove bytes from the document: `ourPatchBytes = consumed` represents the budget we re-budgeted for that retry, but the patch the initial write embedded is still there and still ours. Subtracting the rejected retry bytes understated our ownership, shrank `tightened` on the next pass, and could drive `ourPatchBytes` to 0 so the loop bailed out at the guard with the doc still over the ceiling.

Reconciliation now runs for the **initial write only** — where a rejected op truly never contributed bytes — and the `ourPatchBytes = consumed` retry assignment stays as-is. The comment on the helper and the dropped block both document why.

Validation: `npx jest libs/platformData` → 7 suites, 131 passed (unchanged set — the initial-write reconciliation test still covers the intended behavior). Head confirmed `121c331b9`, mergeable.

Wellington01 (approved):

PR #1902, opened 2026-09-10
fix(workflow): reclaim stranded PROCESSING jobs via lease

### LINKED ISSUE #1830: Workflow jobs stranded in PROCESSING when a worker dies mid-job — in-process timeout cannot recover them
## Summary

Code review jobs get stranded in `PROCESSING` forever when the worker executing them is killed mid-job. The 105-minute abort timeout does not help, because it is an in-process `setTimeout` that dies with the process. The stale-job reaper eventually marks these jobs `PERMANENT` — no retry, no notification — and fires an incident alert whose stated cause is wrong.

This is the same failure class as the August incident (worker killed by the nightly scale-down leaving jobs stranded). The trigger has changed — it is now container health-check failures throughout the day — but the underlying gap is identical: **job ownership has no liveness signal.**

## Impact

- Code reviews silently never complete. The reaper marks them `PERMANENT`, so they are never retried and the requester is never told.
- Incident alerts fire with a misleading cause, costing investigation time on every occurrence.
- The BYOK concurrency gate amplifies the damage: a killed worker holds its slot until the Postgres session drops, so scarce slots (3–6 per configuration) are consumed by dead work.

Observed escalation over four days in production:

| Day | Gate events | `forcing acquisition` | reclaim | reap |
|---|---:|---:|---:|---:|
| Aug 29 | 2,169 | 184 | 2 | 3 |
| Aug 30 | 2,680 | 355 | 3 | 2 |
| Aug 31 | 6,463 | 535 | 22 | 17 |
| Sep 1 | 30,777 | 10,001 | 39 | 24 |

## Evidence

A single code review job, marked `PROCESSING` at 21:00:32 UTC, produced exactly six log lines and then nothing for 2h35m:

```
20:55:31.686  Processing Code Review Job <redacted>
20:55:31.697  [BYOK-CONCURRENCY-GATE] max deferrals exceeded, forcing acquisition
20:55:31.716  [BYOK-CONCURRENCY-GATE] deferred workflow job via outbox relay
21:00:32.140  Processing Code Review Job <redacted>
21:00:32.148  [BYOK-CONCURRENCY-GATE] acquired distributed BYOK slot
              ---- 2h35m of complete silence ----
21:15:01.404  WorkflowJobConsumer: Message already claimed by another worker
```

No pipeline stage, no LLM call, no error, no timeout, no completion. The worker task that owned the job was stopped by ECS at 21:04:59 — 4.5 minutes after it acquired the slot. Its log stream ends at 21:15:00.

The ECS service event history shows **27 task replacements due to `unhealthy status`**. One task was measured freezing for **106.4 seconds** with no log output at all; the container health check probes `/health` every 30s with a 4s timeout and tolerates 3 failures, so a freeze of that length guarantees a kill.

## Root cause

`runWithTimeout` (`libs/core/workflow/infrastructure/run-with-timeout.ts`) schedules the deadline in-process:

```ts
timeoutId = setTimeout(() => {
  controller.abort();
  reject(new Error(timeoutMessage));
}, timeoutMs);
```

With `CODE_REVIEW_PROCESS_TIMEOUT_MS = 105 * 60 * 1000`, this correctly bounds a *slow* job. It cannot bound a *dead* one: when the process is killed, the timer dies with it — no abort, no catch, no `finally`, no transition to `FAILED`. The row stays `PROCESSING` indefinitely.

The reaper then finds it only by age (180 minutes) and cannot tell "the worker died" from "the job is legitimately still running", because there is no liveness signal to consult. That is why the alert's `possibleCause` is a hardcoded guess.

### 1. Heartbeat lease on job ownership (root cause)

Add to `workflow_jobs`:

- `leaseOwner` — identifier of the process holding the job
- `leaseExpiresAt` — timestamp

Behaviour:

- The worker renews `leaseExpiresAt = now() + 90s` every ~30s while processing.
- The reaper reclaims by `leaseExpiresAt < now()`, not by age.
- A reclaimed job returns to `PENDING` with `retryCount++` instead of being marked `PERMANENT`.

This reduces recovery from 180 minutes to roughly 90 seconds, makes the loss recoverable instead of silent, and — importantly — makes the alert meaningful, since an expired lease is evidence of death rather than a guess.

This works regardless of *why* the process died, which is what makes it the priority.

### 2. Make the freeze diagnosable

Instrument the event loop with `perf_hooks.monitorEventLoopDelay`: log when lag crosses a threshold and expose the current value from the `/health` endpoint. The 106s freeze is currently pure silence; nothing in CloudWatch can explain it.

Deliberately **not** proposing to relax the health check (more retries / longer `startPeriod`) — that would hide the symptom. With a lease in place, a kill stops being expensive.

### 3. Fix the alert

- Replace the hardcoded `possibleCause` ("Worker crashes/evictions (OOM, node pressure, restarts)") with real context: time in `PROCESSING`, last known stage, and whether the owning lease is still valid.
- Split `stale_job_reap_rate` and `inbox_reclaim_rate` — they currently fire on the same heartbeat URL, so neither can be routed or silenced independently.
- Revisit the `> 5` threshold. It is tuned for crash-loop spikes and hid a chronic one-job-per-night loss for two weeks in August. Alert on rate against a baseline, and route slow leaks to a low-severity channel instead of leaving them invisible.

## Related improvements

Independent of the above, found during the same investigation:

- **BYOK slot has no TTL on the normal acquisition path.** Only the forced path passes `{ ttl: 30_000 }` — which is precisely why it is the only path that still works when everything is stuck. Without a TTL the slot is released only when the Postgres session drops.
- **`lastError` is not cleared on transition to `PROCESSING`.** `deferJob` writes `status: PENDING` with "Waiting for a BYOK concurrency slot…"; when the job later acquires a slot, the stale message remains and reads as current state. This actively misled the investigation.
- **`currentStage` is `null` on every stranded job.** The column exists but is never populated during execution, so there is no way to tell where time is being spent.
- **Bitbucket 429 backoff is a flat 10s, 4 attempts, no jitter.** The `Retry-After` header *is* honoured, but Bitbucket omits it on per-endpoint burst 429s, so every wait falls back to the fixed value. Workers then retry in lockstep and reinforce the limit. The gate also parks per credential while the 429s are per endpoint.
- **Recursive CTE over `ast_edges`** shows p90 6.6s and max 21.7s. It does not saturate the database (~2.6% utilisation) but it does hold the job — and therefore the slot.

## Ruled out with measurement

For the record, so nobody re-investigates these:

- **Connection pool** — 1 timeout in 22h across API and worker. The August fix holds.
- **Database saturation** — Performance Insights `db.load` max 0.39 over 2h; CPU 18–20%; CPU credits never depleted; connections at roughly 46% of the limit.
- **Outbox backlog** — `READY` rows only from the last few minutes; max `attempts` = 1.
- **CPU / memory on workers** — ~1.4 GB of 10 GB allocated; each task effectively alone on a 4-vCPU instance, with idle instances available.
- **Deploy** — no deployment for 12 days before onset; all services on the same image revision.
- **Load** — job volume *fell* in the hour the escalation began.

andyst-dev:
Fixes #1830

## Problem
A code-review job stays `PROCESSING` forever when its worker is SIGKILLed (OOM, ECS `unhealthy` eviction, nightly scale-down). The 105-min abort timeout is an **in-process `setTimeout` that dies with the process**; the stale-job reaper only knows `updatedAt` age (180 min) and cannot tell a dead worker from a legitimately long run, so it permanently `FAILED` the job — no retry, no notification — and fired an incident alert whose `possibleCause` is a hardcoded guess. Production shows up to **30,777 BYOK gate events/day** and `forcing acquisition` bursts from dead workers holding slots.

## Fix — lease on job ownership (the issue's priority #1)
The reaper now reclaims by **liveness**, not by age:

- **`workflow_jobs` gains `leaseOwner` + `leaseExpiresAt`** (model + `2026091200000000-AddWorkflowJobLease` migration).
- **`CodeReviewJobProcessorService`** acquires the lease on claim (`leaseOwner = hostname`, `leaseExpiresAt = now+90s`) and **renews it every ~30s** for as long as the review runs, via a new `startJobLeaseRenewal` helper (interval `unref`'d so it doesn't hold the worker, stops on completion or parent abort, surfaces renew failures). A dead worker can no longer renew, so its lease lapses in ~90s.
- **Reaper** (`reapStaleProcessingJobs`) now uses `findStaleProcessing`: select PROCESSING rows whose lease **expired** (`leaseExpiresAt < now` — a renewed lease means the worker is alive and the job is left alone) or, for legacy pre-lease rows, `updatedAt` older than the cutoff. Within the retry budget it **requeues to `PENDING` with `retryCount++`** (clearing lease/stage/startedAt) instead of permanently failing; only retry-budget-exhausted jobs are `FAILED`/`PERMANENT`. The alert cause is now **evidence** (lease expired), not a guess.

Net: recovery drops from **180 min → ~90 s**, the job is **recoverable** (re-queued for retry) instead of silently `PERMANENT`, and the incident alert names the real cause.

## Validation
- `job-lease.renewer.spec.ts` — renewal cadence; stops on `stop()`/abort; async failure surfaced via `onRenewError` without crashing.
- `workflow-job.repository.spec.ts` — `findStaleProcessing` lease-vs-age predicate; `requeueStaleJobs` → PENDING + retry++ + lease cleared; empty-list no-op; `failStaleJobs` → FAILED/PERMANENT.
- `outbox-relay.service.spec.ts` — reclaimed-for-retry (not permanently failed) under budget; permanently failed at maxRetries; cutoff override; high-reap-rate incident; lock release on failure.
- **20/20** across the three specs; `tsc` clean on all touched files; `eslint` clean.

## Scope left (explicit)
- **Re-driving a reclaimed `PENDING` job still needs a fresh trigger** (a later webhook redelivery for that PR). Automatic re-publish per workflowType from the reaper is a follow-up — I've kept the reclaim at the DB level so it's reviewable and safe.
- The **event-loop-lag / `/health`** freeze diagnostic (issue part #2) and the **RabbitMQ 30-min ack re-loop** are separate and intentionally out of scope here.


andyst-dev:
Addressed the review. Lease renewal failures now go through the logger instead of only the optional callback, and after a couple of consecutive failures (default 2) the run is aborted so a worker cannot keep executing an unleased job. The requeue partition uses a Set, the retry predicate is now retryCount < maxRetries (so maxRetries:1 jobs actually requeue), and requeue plus dead-letter run under Promise.allSettled. On the SQL side the stale-lease disjunction is parenthesised so the PROCESSING guard covers both branches, both reaper UPDATEs re-assert status = PROCESSING, and findStaleProcessing quotes and aliases its columns. One note: there is no PinoLoggerService in this repo, the canonical logger is SimpleLogger via createLogger, so I logged through that. 56 tests pass.

andyst-dev:
Addressed both from the second round. Pushed in `d80744466`.

1. **Lease loss made the job terminal.** `onLeaseLost` aborts the run, but that rejection surfaced inside the processor's `try`, so the `catch` still ran `handleFailure` (status FAILED, PERMANENT) and `notifyReviewFailed`. Since the reaper only lists PROCESSING jobs, a transient renewal failure (pool exhaustion, lock timeout) would permanently kill a job the PR exists to rescue. The processor now records the abort as lease loss in a flag and, in the `catch`, skips the terminal failure and the author notification for that case, leaving the row PROCESSING so the expired lease is reclaimed and requeued. Note: the flag is declared in `process` scope rather than inside the `try`, because a `catch` is a sibling lexical scope and would not see a `let` from the `try` block.

2. **organizationId missing on the requeue-failure log.** Added the affected `organizationIds` (deduped from the requeueable batch) to the metadata of the "Failed to requeue stale PROCESSING workflow jobs" error, so a batch reclaim failure is attributable to its tenants.

Validation:

- `SKIP_INTEGRATION=true npx jest libs/core/workflow libs/code-review/workflow` -> 15 suites passed, 147 passed, 8 skipped (the skipped ones are the Postgres integration specs).
- The lease-lost spec now also asserts the negative: no FAILED status write and no `notifyReviewFailed` call on a lease-lost abort. It failed before the fix and passes after.
- `tsc -p tsconfig.json` clean on the touched files (the remaining tsc output is pre-existing on main).
- PR still reads `mergeable: true`.

andyst-dev:
Good call, and you're right that my throw defeated the purpose. Throwing let `JobProcessorRouterService`'s catch (and the consumer's) write `FAILED/PERMANENT`, and the lease reaper only reclaims rows still in PROCESSING — so the advertised reclaim never happened, and it'd be a silent permanent failure with no author notice.

Fix in `bbda8bc3a`: the lease-lost branch now logs a warning and **returns** instead of throwing. Resolving the processor lets the consumer ack the message with the row left untouched in PROCESSING, so the reaper requeues it once the lease expires — the behavior #1830 intended.

Validation: `SKIP_INTEGRATION=true npx jest libs/code-review/workflow` → 7 suites, 85 passed. The lease-lost spec now asserts `resolves` (not `rejects`) plus no FAILED write and no `notifyReviewFailed` call; I confirmed it's RED against the old throw path (it fails) and green now. Head confirmed `bbda8bc3a`, `mergeable: true`.

Thanks for tracing it through the router/consumer.

andyst-dev:
Good catches — all three addressed in `f7ad6aac1`.

1. **orgId on the lease-lost warning.** The salvage-path warn now carries `organizationId` (read from the job payload, since the `organizationAndTeamData` const in the `try` isn't visible to the `catch`).

2. **The reclaimed job was never re-published — a real gap.** `requeueStaleJobs` only flips the row to PENDING; since the consumer only re-runs jobs that arrive on `workflow.jobs.*`, the reclaimed job sat dead until an external trigger. Fixed properly:
   - `requeueStaleJobs` now returns the UUIDs it *actually* flipped, via Postgres `RETURNING` in the guarded UPDATE. This is the TOCTOU-safe part: a job that finished between the SELECT and the write is excluded, so it is never re-driven.
   - `outbox-relay` publishes `workflow.jobs.resumed.<workflowType>` for exactly those requeued UUIDs, with a fresh `reclaim-<uuid>-<ts>` messageId so the inbox claim isn't a no-op — mirroring the `workflow.jobs.resumed` pattern the AST/heavy-stage resume paths already use.
   - I verified the consumer binds `workflow.jobs.*.CODE_REVIEW` (wildcard middle segment), so a `resumed` routing key is consumed correctly.

3. Skipped-verification/log consistency carries through elsewhere.

Validation: `jest libs/core/workflow libs/code-review/workflow` → 15 suites passed, 149 passed, 8 integration skipped. Two new tests cover the republish path (per-job routing key + fresh messageId) and the TOCTOU guard (no republish for UUIDs the guarded UPDATE did not return); I confirmed the republish test is RED without the block. tsc clean on all touched files. Head confirmed `f7ad6aac1`, mergeable.

andyst-dev:
All three are fair, fixed in `f27da8d`.

**1. The publish loop was fail-fast, and that defeated the fix.** `await Promise.all(...)` on the republishes means a single broker rejection (`messageBroker` throws `RabbitMQ connection is not available` when the broker is down) propagates out of the block into the reaper's `catch`. The DB requeue has already committed at that point, so the jobs whose message never went out stay PENDING with no broker message — exactly the #1902 failure this block was added for — and the remaining publishes plus the cycle's heartbeat/reap report are skipped. The publishes now go through `Promise.allSettled`, with per-job failure logging (uuids, orgs, reasons) instead of aborting the batch. Each publish also moved into an `async` private method, so even a client that throws synchronously is confined to its own job rather than breaking out of the `.map()` that builds the list.

**2. `includes` inside the filter.** Replaced with a `Set` built once from `requeueResult.value`, so the `typeByUuid` construction is O(stale + requeued) per tick instead of O(stale x requeued).

**3. Missing organizationId.** The batch's org ids (deduplicated) are now threaded into `requeueStaleJobs` and logged in its metadata next to the uuids, so a reclaim can be filtered per tenant in the log system. I passed them from the caller rather than resolving them from the requeued rows: the repo's `RETURNING 'uuid'` keeps telling us exactly which rows flipped, and the union of the batch's orgs is the traceable set. The same ids go on the republish-failure log.

Validation: `outbox-relay.service.spec.ts` + `workflow-job.repository.spec.ts` are 26/26. Three new cases: org ids forwarded to the repo call, an async publish rejection still completing the cycle (6 reclaimed > high-reap threshold, so the incident heartbeat is reached), and a synchronous broker throw isolated to its own job. RED verified on the pre-fix source: 3 failed / 23 passed, only the new cases failing. eslint and prettier clean on all four files.

libs/core/workflow/infrastructure/repositories/workflow-job.repository.ts
```diff
@@ -318,6 +323,176 @@ export class WorkflowJobRepository implements IWorkflowJobRepository {
...
+                .select([
+                    'job."uuid"',
+                    'job."workflowType"',
+                    'job."organizationId"',
+                    'job."startedAt"',
+                    'job."leaseExpiresAt"',
+                    'job."retryCount"',
+                    'job."maxRetries"',
+                ])
+                .from(WorkflowJobModel, 'job')
```
malinosqui, line 352: `createQueryBuilder()` without an alias already selects FROM `workflow_jobs` as `WorkflowJobModel`, so `.from(WorkflowJobModel, 'job')` adds a second FROM and turns this into a cross join. The generated SQL is in the review summary. Pass the alias to `createQueryBuilder` and drop `.from`:

```ts
const result = await this.repository
    .createQueryBuilder('job')
    .select([ /* same columns */ ])
    .where('job.status = :status', { status: JobStatus.PROCESSING })
    // ...
```

libs/code-review/workflow/code-review-job-processor.service.ts
```diff
@@ -208,9 +226,78 @@ export class CodeReviewJobProcessorService implements IJobProcessorService {
...
+            // runs. A killed worker (OOM, ECS eviction, kill -9) cannot renew,
+            // so its lease lapses and the reaper reclaims instead of waiting
+            // out the 180-min in-process timeout that dies with the process.
+            const leaseRenewal = startJobLeaseRenewal({
+                signal: runSignal,
+                logger: this.logger,
+                jobId,
+                organizationId: organizationAndTeamData?.organizationId,
+                renew: () =>
+                    this.jobRepository.update(jobId, {
```
malinosqui, line 263: This renewal is unconditional. If the reaper has already reclaimed the job and another worker owns it, this write silently takes the lease back, and both workers run the review. Scope it to `uuid = :id AND "leaseOwner" = :instanceId AND status = 'PROCESSING'`, and treat 0 affected rows as lease lost (abort). The same guard belongs on `markCompleted`.

malinosqui (changes requested):
Thanks for taking this on! A liveness lease is the right fix for #1830, and moving the reaper from "fail after 180 min" to "reclaim and retry" is a big improvement. Before merging, three things need work: one blocking bug, one correctness gap, and one behavior change that needs an explicit decision.

### 1. Blocking: `findStaleProcessing` cross-joins `workflow_jobs` with itself

`this.repository.createQueryBuilder()` already puts the entity in `FROM` under the default alias `WorkflowJobModel`. `.from(WorkflowJobModel, 'job')` then adds a second `FROM` entry. I generated the SQL with the repo's TypeORM (same builder chain, no DB connection):

```sql
SELECT job."uuid", job."workflowType" FROM "workflow_jobs" "WorkflowJobModel", "workflow_jobs" "job"
WHERE "job"."status" = :status AND ((job."leaseExpiresAt" IS NOT NULL AND job."leaseExpiresAt" < :now) OR (job."leaseExpiresAt" IS NULL AND job."updatedAt" < :olderThan))
```

Nothing constrains the `WorkflowJobModel` alias, so every stale job is returned once per row in `workflow_jobs`, and nothing prunes that table. In production, the reaper would load (stale jobs × table size) rows every run and pass that many uuids to `requeueStaleJobs`. The resulting `IN (...)` goes past Postgres' 65,535 bind-parameter limit, the requeue is rejected, and nothing gets reclaimed, which is exactly the failure this PR is meant to fix.

The repository specs mock the query builder, so they can't catch this. Please:
- use `this.repository.createQueryBuilder('job')` and drop `.from(...)` (suggestion inline);
- add an integration test that runs `findStaleProcessing` and `requeueStaleJobs` against a real Postgres, with a few non-stale rows in the table, and asserts each stale job comes back exactly once. `test/integration/parameters/create-or-update-config-race.integration.spec.ts` shows the real-Postgres setup. Note that fork CI skips migrations, so please run it locally and paste the output.

### 2. No fencing: a frozen worker can double-run a job

The renewal is an unconditional `jobRepository.update(jobId, { leaseOwner, leaseExpiresAt })`, and `markCompleted` doesn't check ownership either. #1830 documents a worker frozen for 106s without dying. Here is how that plays out:

1. Worker A freezes for more than 90s, so its lease expires.
2. The reaper requeues the job to `PENDING` and republishes it, and worker B picks it up.
3. Worker A unfreezes. Its next renewal overwrites B's `leaseOwner`/`leaseExpiresAt`, and A runs to completion.
4. Both A and B post the review, so the PR gets duplicate comments.

Please make both the renewal and the terminal transitions conditional on ownership, e.g. `UPDATE ... WHERE uuid = :id AND "leaseOwner" = :me AND status = 'PROCESSING'`. If a renewal affects 0 rows, treat it as lease lost and abort the run, the same path as the 2-consecutive-failures case you already have.

### 3. Behavior change for workflow types that don't take a lease

Only `CodeReviewJobProcessorService` acquires a lease. Every other workflow type (no lease, so `leaseExpiresAt IS NULL`) hits the legacy age branch. On `main` those rows are failed as `PERMANENT`. With this PR they are requeued and republished up to `maxRetries` times. For a job that kills its worker (e.g. an OOM on a huge repo), that means taking down several more workers before it finally fails. This may well be what we want, but it should be a deliberate choice. Options: keep the legacy branch as fail-only, limit requeueing to rows that actually had a lease, or call the change out explicitly in the PR description.

Minor: the PR description still says "Re-driving a reclaimed PENDING job still needs a fresh trigger", but the reaper now republishes reclaimed jobs. Please update the description.

PR #1912, opened 2026-09-13
fix(code-review): gate git-issue refs on a git-issues MCP (#1908)

### LINKED ISSUE #1908: Business-logic gate accepts GitHub `#N` refs with no git-issues MCP connected, then posts "Insufficient Task Context"
### Issue Summary

Business-logic validation runs on a PR whose only task reference is a git-issue ref (`#1868`, `Closes #1868`) even when **no git-issues MCP is connected** — as long as the org has *any* MCP from `TICKET_KEY_MCPS` attached (Jira, Linear, ClickUp, Atlassian Rovo...). The skip gate checks that a ticket key exists and that *some* ticket-key-capable MCP is connected, but never checks that the connected MCP can resolve *that kind* of reference.

The run then has no tool that can read the issue, flails through the Jira tools, latches onto whatever the generic Atlassian `search` returns, classifies it `MINIMAL` and posts **"🤔 Insufficient Task Context"** on the PR — on a PR whose linked issue has a full description.

This is our own `kodustech/kodus-ai` org today: it tracks work in GitHub issues, and the only task-management MCP connected is **Atlassian Rovo**. Every `fix(...)` PR that closes a GitHub issue gets the canned comment.

### Steps to Reproduce

1. Org has a task-management MCP connected that is in `TICKET_KEY_MCPS` but is **not** git-issues (e.g. Atlassian Rovo / Jira).
2. Org does **not** have the Kodus "Git Issues" MCP connected.
3. `business_logic` review option is on.
4. Open a PR whose body contains `Closes #1868` (a GitHub issue), with no Jira key anywhere in body/title/branch.
5. Let the review run.

### Actual Results

Kody posts:

> 🤔 **Insufficient Task Context** — I found a task linked to this PR, but it only contains minimal information (title only, no description or acceptance criteria).

No LLM call is ever made — the blueprint short-circuits at the `validateTaskContext` gate. Reproduced on PR #1861 (6 times), and still happening on #1904 and #1907.

### Expected Results

Either:

- **skip silently** with `reason: no_signals`, because `#1868` cannot be resolved by any connected MCP — no PR comment at all; or
- resolve the reference through a git-issues MCP when one is attached.

What must not happen: telling the user their task description is insufficient when the product never read the task, and the issue in question has ~3 KB of description.

### Technical details

Root cause is one gate, `libs/code-review/pipeline/stages/business-logic-validation.stage.ts`:

```ts
// hasRelevantBusinessSignals()
const ticketKeys = this.detectTicketKeys(body);   // -> ['#1868']
if (ticketKeys.length > 0 &&
    connectedMcps.some((mcp) => TICKET_KEY_MCPS.includes(mcp))) {
    return true;                                   // 'atlassianrovo' satisfies this
}
```

`detectTicketKeys` deliberately pushes `#N` refs into the same bucket as `ABC-123` Jira keys, and the membership test is "any ticket-key MCP", not "an MCP that can resolve this reference". The dedicated branch right below it — the one that requires `gitissues`/`githubissues` — is unreachable for `#N`, because the first branch already returned `true`.

Downstream, nothing recovers:

- `matchesRequestedTask` (`libs/agents/skills/capabilities/task-context/task-identity.ts`) only rejects on *positive* evidence of a different task. An Atlassian search hit carries no number or key comparable to `#1868`, so `comparable === false` and the bogus hit is accepted as "the task linked to this PR".
- `classifyTaskQualityFromSources` sees a title with no description → `MINIMAL` → `taskContextStatus: 'weak'` → gate short-circuits → canned message.

### Evidence

Prod worker logs, `/ecs/kodus-prod-orchestrator/worker`, PR #1861, 2026-09-06 13:37 UTC (`corr_iZFXPPH4XiEh_mtpuwfex`):

```
[BUSINESS-LOGIC] running business-rules agent (passed skip gate)
  signals: { ticketKeys: ["#1868"], taskLinks: [], requirementKeywords: ["when","then"] }

[GenericSkillRunner] MCP adapter created for skill 'business-rules-validation'
  servers: [
    { name: "Kodus MCP",      allowedTools: ["KODUS_GET_PULL_REQUEST","KODUS_GET_PULL_REQUEST_DIFF"] },
    { name: "Atlassian Rovo", allowedTools: ["atlassianUserInfo","getAccessibleAtlassianResources",
                                             "getJiraIssue","getTransitionsForJiraIssue", ...,
                                             "searchJiraIssuesUsingJql","search","fetch"] }
  ]
  registeredToolCount: 15
```

None of those 15 tools can read a GitHub issue — `KODUS_GET_ISSUE` is not among them, because the Git Issues MCP is not connected.

Tool attempts for that run:

```
getJiraIssue             failed  401 "Unauthorized; scope does not match"
searchJiraIssuesUsingJql failed  400 "O token apresentado na próxima página é inválido ou expirou"
search                   success (Rovo generic search)
```

Blueprint outcome:

```
[blueprint] gate 'validateTaskContext' failed — short-circuiting
completedSteps: ["fetchPullRequestMetadata","fetchTaskContextFromMcp","classifyTaskContext"]
skippedAt: "validateTaskContext"
```

Langfuse confirms no model ever ran: session `04bd288b-595a-4ee1-87cd-8bbbdc312b3c:956679248:1861` has 14 traces, all `kodus-generalist-review-agent` / `kodus-rules-review-agent`, and **zero** `business-rules-validation`. The whole stage took 3.7 s.

Control case, same org, same day, same broken Rovo connection: PR #1869 had `KC-1559` in the body. `getJiraIssue` and the JQL search failed there too (9 attempts each), but the generic `search` landed on the right Jira issue *with a description* → `TASK_QUALITY: PARTIAL` → the analyzer actually ran (Langfuse trace `1dd834cdc5a0829bd3405e160d870036`). So the failure is specific to references the connected MCP cannot resolve.

Still reproducing after the #1812 fix: PR #1904 (2026-09-11 12:06 UTC) and PR #1907 (2026-09-11 18:40 UTC), same three capability traces.

Not verified: the exact item Rovo's `search` returned — the tool result is not logged. `MINIMAL` means it had a title and no description (or a stub under 80 chars), which is all we can say.

### Relationship to existing issues

- **#1884** (open) names this exact case — *"ticket-key detection is not source-scoped (`AB#5625` or `#256` can satisfy the skip gate when only Jira is connected)"* — but puts it **out of v1**: *"Source-scope `#N` vs Jira on the 'managed' / unset path: out of v1. Only a selected `integrationId` source-scopes detectors."* So shipping #1884 as specced leaves this org in the same state.
- **#1812** (closed as completed 2026-09-05) listed *"The task/issue key validation doesn't always work correctly"* as its third problem. PR #1840 fixed the one-shot repetition (problems 1 and 2); the third went out with the close.
- **#1810** (open) is the self-hosted Linear variant of the same mismatch.

This issue is deliberately narrow: make the skip gate match the *kind* of reference against the *kind* of connected MCP, so a PR with only `#N` refs and no git-issues MCP skips silently instead of posting a wrong diagnosis. It does not need the #1884 architecture and can ship before it.

### Side note, not this issue

`getJiraIssue` and `searchJiraIssuesUsingJql` fail on **every** run for this org (2026-09-06 and 2026-09-11, both PRs), with 401 scope-mismatch and 400 responses. The generic `search` is carrying the whole capability. Worth its own look.

andyst-dev:
## Summary

Fixes #1908

The business-logic skip gate treated any git-issue reference (`#N`, or a full `github.com/org/repo/issues/N` URL) as a valid signal for **any** connected ticket-key MCP. When the org's only task-management MCP was Atlassian Rovo (which cannot resolve git-issue refs), the gate passed, the agent flailed through the wrong tools, and it posted "🤔 Insufficient Task Context" on a PR whose linked issue had a complete description.

Now a git-issue reference only counts as a signal when a git-issues MCP (`gitissues` / `githubissues`) is actually connected. Jira-style keys (`ABC-123`, `PROJ_1-42`) keep matching any ticket-key MCP exactly as before — mixed cases (a Jira key alongside a git ref) still trigger validation.

## Test plan

- [x] `SKIP_INTEGRATION=true jest test/unit/code-review/pipeline/stages/business-logic-validation.stage.spec.ts` → 38 passed (6 new `hasRelevantBusinessSignals` cases + 2 end-to-end `evaluateSkip` cases covering the #1908 scenario).
- [x] RED verified: with the old signal logic restored, 3 of the new tests fail; they pass with the fix.
- [x] `tsc --noEmit` clean on touched files; eslint clean (the one `prBody` unused warning at `detectSignals` line 446 is pre-existing on main, outside this diff).

## Notes for the reviewer

The behaviour change is strictly a tightening: a PR whose only reference is a git issue and whose org has no git-issues MCP now skips business-logic validation cleanly (`no_signals`) instead of running and misreporting. No legitimate case regresses — a real Jira/Linear/Rovo key still triggers validation whenever a matching MCP is connected, and git issues still validate when a git-issues MCP exists.


andyst-dev:
Good call — the bare `/issues/\d+` regex was too loose. It also matched fixture paths, branch names like `fix/issues/191`, and prose like "see issues/2024", which would let the business-logic run against a PR with no resolvable issue reference (the very gap #1908 closes).

Fix in `0b77931ab`: the git-issue URL detector is now anchored to a real URL — `https?://[^\s)>]"']*/issues/\d+` — mirroring the scheme-anchored pattern `detectTicketKeys` already uses. A bare `#N` is still matched by the first branch.

Validation: `jest test/unit/code-review/pipeline/stages/business-logic-validation.stage.spec.ts` → 39 passed (added regression cases: `see issues/2024` and `test/fixtures/issues/123.json` both return false with a git-issues MCP connected; the full `https://github.com/.../issues/N` still returns true). Head confirmed `0b77931ab`.

andyst-dev:
Good catch, fixed in `f8bf720`.

The URL form had lost the `i` flag its sibling `detectTicketKeys` pattern carries (`/...\/issues\/(\d+)/gi`), so `HTTPS://github.com/acme/proj/issues/1825` no longer set the signal. On a git-issues-only MCP setup that is not a cosmetic miss: `#N` keys are filtered out of the ticket-key set in this same function and `detectTaskLinks` is case-sensitive too, so the stage returned `no_signals` and silently skipped business-logic validation for a PR that does link a resolvable issue. The regex now carries the flag, mirroring `detectTicketKeys`.

Regression test added for `HTTPS://` and `Https://` references with a git-issues MCP connected. RED verified against the pre-fix regex (the new case was the only failure: 39 passed / 1 failed on the old code), GREEN with the flag: `business-logic-validation.stage.spec.ts` 40/40. eslint reports only the pre-existing `prBody` warning inherited from main, untouched by this diff.

Issue #1826, opened 2026-09-01
Kody Rules: let a rule declare the context it needs beyond the diff


malinosqui:
## Problem

Some Kody Rules can only be judged with information that is not in the diff. Today every rule is judged as if the diff were the whole world. That produces two failure modes:

- **False positives that break trust.** A file-scope "remove unused imports" rule flagged an SCSS import that was used outside the changed hunk (#1724). Removing it would have broken the build. The evaluator could not tell a hallucination from missing context, which is worse than either alone.
- **False negatives that make a whole class of rule useless.** Rules like "every new endpoint has a test" or "use the shared logger instead of console.log" cannot fire correctly, because the judge has no way to know whether the test exists or where the shared logger lives.

The product claim is that rules can read the rest of the repository when they need to. For the Kody Rules path that claim does not hold, and the code still carries dead prompt text saying "use readFile/grep to verify" that is never reached (`kody-rules-agent.provider.ts`, `getCategoryPrompt` / `buildUserPrompt`).

## Why we moved to a diff-only judge

Before #1449, Kody Rules ran through the same agentic loop as the generalist agents: one agent per PR, all rules in context, tools available, and a turn budget. The model decided which files to open. On large PRs it spent the budget and never opened the file with the violation. Measured on the frozen `github-cases` set: **40% occurrence-recall on gpt-5.4, 58% on kimi**.

We replaced traversal with a deterministic file x rule sweep (`kody-rules-sharded.judge.ts`):

- one single-shot structured call per changed file, with the rules whose `path` matches;
- one whole-PR call for `pull-request` scope rules, with title, description, file list and the full diff of every changed file (150k char budget);
- compiled T0 detectors run as regex over added lines, no LLM at all.

Coverage became structural: the model only judges, it never decides where to look. Recall went to **91 to 100% across gpt-5.4, gpt-5.4-mini and kimi at the same or lower cost**, and weak models gained the most. That win is real and we should not give it back.

The cost of that design is that neither shard can see anything outside the diff:

| The judge sees | `file` rule | `pull-request` rule |
|---|---|---|
| Diff of the changed file(s) | its own file only | all files in the PR |
| PR title, body, file list | no | yes |
| Unchanged lines of the same file | no | no |
| Files in the repo not touched by the PR | no | no |
| `grep` / `readFile` / any lookup | no | no |

[...]

PR #1914, opened 2026-09-14
fix(kody-rules): scope library rules by language when path empty (#1832)

### LINKED ISSUE #1832: Kody Rules library import drops rule language and stores empty path, so rules apply to every file
## Summary

Importing a rule from the Kody Rules library drops the library entry's `language` and stores the rule with an empty `path`. The engine's only scoping mechanism is `rule.path` (`if (!rule.path) return true`), so a Ruby-specific library rule lands in the org applying to **every file in every PR**.

## Evidence

- Library entry: `{ "title": "Avoid Using Semicolons to Terminate Statements", "language": "ruby", "scope": "file" }` (no `path`).
- After import in production: `path: ""`, `scope: file`, `origin: library`.
- Base-wide (BigQuery, active rules): 9,227 rules, **4,697 (51%) with no path**; of the 2,654 library-origin rules, **1,763 have no path**.
- Consequence measured on one customer: the Ruby semicolon rule fired 349/365 times on non-Ruby files (see the T0 detector issue); rules like "N+1", "Memory Bloat", "Add error handling for API response access" fired on spec files (24% of all rule thumbs-down are on test/spec/fixture files, vs 2% for the generalist).

## Where

- `libs/kodyRules/application/use-cases/add-library-kody-rules.use-case.ts:58,97` — copies `path: libraryKodyRules.path` (undefined) and never reads `language`.
- `libs/code-review/infrastructure/agents/providers/kody-rules-agent.provider.ts` — `applicableRules` filter: empty path ⇒ applies to all files.
- Rule editor UI: `path` optional, no hint when the rule text names a language.

## Proposed fix

1. On import, map the library `language` to a path glob (`ruby` → `**/*.{rb,rake,erb}` etc.) and persist it; keep `language` on the rule for the detector compiler.
2. In the rule editor, when the rule text names a language and `path` is empty, suggest the glob; add an explicit "applies to test files?" choice that writes into the rule text/path.
3. Backfill the 1,763 library rules with a derivable language; report the rest for CS.
4. (Optional) pass "this file is a test file" to the judge so production rules are only applied to tests when the rule evidently targets tests.

## Impact

Root cause behind most RULE_OVERAPPLIED feedback (27% of a 1,252-item classified sample of thumbs-down; ~35% population-weighted).

## How we found this

We ran a structured analysis of every thumbs-down (👎) reaction left on Kody suggestions over four weeks (2026-08-04 → 2026-09-01), using the BigQuery mirror of production (`codeReviewFeedback` joined to the suggestion bodies in `pullRequests`):

- 2,515 👎 vs 1,915 👍 across 65 organizations; 3.8–5.6% of delivered suggestions get a 👎 in a given ISO week.
- 66% of all 👎 are Kody Rules suggestions (1,670), 34% generalist (845). One organization accounts for 43% of the total.
- We classified 1,252 of them one by one (a random 500, then an org-stratified increment capped at ~60 per org so no single customer dominates) into 12 buckets with an LLM judge, and cross-checked every headline number with deterministic queries (file extensions, test-file paths, empty `improvedCode`, detector-shaped summaries).
- We then computed the 👎 rate per **origin** of the suggestion (generalist agent, rules via LLM judge, rules via T0 detector) over the full delivered volume, which is the table above. That table is what turned a "one customer has a bad rule" story into a product defect: the detector path rejects at 25× the generalist rate.
- Finally we traced the mechanism in code (compile gate → `buildDetectorViolations` → publication) and verified the "even scoped to Ruby it is wrong" claim by reading the 16 hits on `.rb`/`.erb` files.

Full write-up with per-bucket and per-org detail is in the internal report "Por que a Kody leva 👎" (2026-09-01).

## Acceptance criteria

- Importing a library rule with `language` set produces a rule with a non-empty, language-appropriate `path` (or an explicit language field the engine honors).
- The rule editor shows a scope suggestion when the rule text names a language and `path` is empty; an explicit "applies to test files" choice exists and is persisted.
- Backfill: library-origin rules without path get a derived scope; a report lists the ones that could not be derived.
- Regression test: the "Avoid unnecessary semicolons" library rule imported into a fresh org does not apply to non-Ruby files.
- Metric: share of active rules without path drops from 51%; rule-judge 👎 on test/spec files drops from 24% of rule 👎.

## Why this is ours

The customer imported a rule from our library. The library knows the language; the import threw it away; the engine has no other scoping mechanism. Writing "does not apply to spec files" in the rule text works for judge-path rules today (the judge reads the rule) but not for detectors, and the customer is not told any of this.

andyst-dev:
## Summary

Fixes #1832

Importing a rule from the Kody Rules library dropped the library entry's `language` and persisted the rule with an empty `path`. The engine's only scoping mechanism is `rule.path` (`if (!rule.path) return true`), so a Ruby-specific library rule landed in the org applying to **every file in every PR** — the root cause behind most RULE_OVERAPPLIED feedback described in the issue.

This closes the import half (acceptance criterion 1):

- `AddLibraryKodyRulesDto` now accepts an optional `language`.
- `AddLibraryKodyRulesUseCase` resolves the path it persists: an explicit `path` wins; otherwise a glob is derived from the declared language (`ruby` → `**/*{.rb,.rake,.erb,.gemspec}`, `python` → `**/*.py`, `typescript` → `**/*{.ts,.tsx}`, …). When neither `path` nor a known `language` is present, the pre-fix behaviour (whatever was given) is preserved so no install changes scope unexpectedly.

Out of scope here (separate, larger): the rule-editor scope suggestion (2), the 1,763-row backfill (3), and passing "this file is a test file" to the judge (4).

## Test plan

- [x] `jest libs/kodyRules/application/use-cases/add-library-kody-rules.use-case.spec.ts` → 4 passed (explicit path wins; single-extension glob; brace glob for multi-extension languages; unknown/absent language keeps current behaviour).
- [x] RED verified: disabling the language-derived glob fails the two glob tests; they pass with the fix.
- [x] `tsc --noEmit` clean on the touched files (the two `results.push` type errors at lines 131/177 are pre-existing on main — this diff does not touch them). `eslint` and `prettier --check` clean.

## Notes for the reviewer

The `LANGUAGE_EXTENSIONS` table mirrors the canonical `SUPPORTED_LANGUAGES` in `libs/code-review/domain/contracts/SupportedLanguages.ts` but is kept local to the use-case to avoid a kody-rules → code-review domain import. If you'd prefer the single source of truth, I can export a `languageToGlob` helper from that module instead. The `#1826` `fileScope` (inferred file-extension scope) is a separate, complementary mechanism and is unaffected.


andyst-dev:
Both points were real, thanks. Fixed in `be02c42`.

**1. Wrong key namespace.** The table was keyed `typescript`/`javascript`, but the library payload carries the `ProgrammingLanguage` keys the web client declares (`apps/web/src/core/enums/programming-language.ts`): `jsts` for JS/TS, plus `dart` and `kotlin`. So the most common imported rules still missed the table, kept `path: ''` and applied to every file in every PR, the exact over-application #1832 is about. The table is now keyed on those 10 keys (`jsts` → `**/*{.js,.jsx,.ts,.tsx}`, `dart` → `**/*.dart`, `kotlin` → `**/*{.kt,.kts}`, ...) and the lookup trims and lowercases the incoming value.

**2. Prototype members were reachable.** `LANGUAGE_EXTENSIONS[language]` was a raw property read, so `'__proto__'` returned `Object.prototype` and threw on `.join` (500 before the rule was created), and `'constructor'` returned the `Object` function and produced the glob `**/*undefined`. The table is a `Map` now, so unknown and prototype-shaped values simply miss and the caller's path is preserved.

Validation: `add-library-kody-rules.use-case.spec.ts` is 7/7, driven by the real client keys plus the prototype-shaped cases. RED verified on the pre-fix source (4/7 failing, including the `.join` TypeError), GREEN on the fix. eslint clean on both files; `tsc -p tsconfig.json` reports the same pre-existing errors before and after (this worktree has no web dependencies).

malinosqui (changes requested):
@andyst-dev 

Blocking: the backend mapping is not exercised by the actual library import flow.

`modal.tsx` builds `newRule` without copying `rule.language`, and `addKodyRuleToRepositories` spreads that object into the request. The endpoint therefore receives `path: ""` with `language: undefined`; `resolveLibraryRulePath("", undefined)` returns `""`, and the engine still treats the rule as applicable to every file.

Please forward the library language in the import payload and add a regression test through the real import boundary. The test should import the Ruby semicolon rule and assert that the persisted path is non-empty and does not match a non-Ruby file.

Directory imports also need the two scopes combined. Today the UI sends `<directory>/**`, and the helper gives any explicit path precedence, so a Ruby rule imported into a directory still applies to every language in that directory. The resulting scope should preserve both the directory and language constraints.

This PR also covers only acceptance criterion 1; editor guidance, test-file selection, and the existing-rule backfill remain open and #1832 should not be closed by this PR.

libs/kodyRules/application/use-cases/add-library-kody-rules.use-case.ts
```diff
@@ -18,6 +18,65 @@ import { CreateOrUpdateKodyRulesUseCase } from './create-or-update.use-case';
...
+ * keys the library payload carries
+ * (`apps/web/src/core/enums/programming-language.ts`: `jsts` covers JS/TS,
+ * plus `dart`/`kotlin`) — those are the values the web client forwards as
+ * `language` when importing a rule from the library. When a rule declares a
+ * `language` but ships no `path`, the engine's only scoping mechanism
+ * (`if (!rule.path) return true`) would otherwise apply the rule to EVERY file
+ * in every PR (#1832).
+ */
+const LANGUAGE_EXTENSIONS = new Map<string, string[]>([
+    ['jsts', ['.js', '.jsx', '.ts', '.tsx']],
```
malinosqui, line 33: Several `jsts` library rules target Vue single-file components and ESM/CJS modules (see review summary).

```suggestion
    ['jsts', ['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts', '.cts', '.vue']],
```

malinosqui (changes requested):
Thanks for this! Using the library's declared `language` is valuable: most library rule texts never name their language (63 of the 65 `ruby` entries don't mention Ruby in the title or rule), so nothing downstream can infer it from the text alone. A couple of issues with the generated globs, plus one design question.

### 1. The `jsts` glob misses Vue and ESM files

These `jsts` library rules target files the glob excludes, so after import they would never fire where they matter:

- Single-file component filename casing
- Single-file component top-level element order
- vue/attribute-hyphenation
- Prefer `const` over `let`/`var` in Vue components
- Disabling Vue.js built-in escaping is security-sensitive
- Use ES Modules (`import`/`export`)

Suggestion inline: add `.mjs`, `.cjs`, `.mts`, `.cts` and `.vue`.

### 2. Manifest-file rules become unreachable

Some rules are about build or manifest files, but the language glob only matches source files:

- `go`: "Use go.work at repo root instead of replace in modules" → scoped to `**/*.go`, never sees `go.work`/`go.mod`
- `rust`: "Centralize Rust deps with workspace.dependencies" → `**/*.rs`, never sees `Cargo.toml`
- `java`: "Gradle: define modules in settings.gradle and use version catalogs" → `**/*.java`, never sees `settings.gradle(.kts)`
- `csharp`: "Central Package Management for .NET" → `**/*.cs`, never sees `.csproj` / `Directory.Packages.props`

Either include the manifest patterns for those languages (`path` accepts comma-separated globs), or skip the language glob for rules like these.

### 3. Question: `path` or `fileScope`?

Since `830d57b3d` (#1826), rules carry a `fileScope` (`IKodyRuleFileScope`: `extensions`, `sourceHash`, `source`) that the review already uses to skip files outside the rule's language (`extensionScopeAppliesToFile` in the sharded judge). A `source: 'author'` scope is never overwritten by inference. Writing the language scope into `path` instead has a few side effects:

- **Visible `path`:** users see a `path` they never wrote, and the settings list truncates brace globs on commas (`showLastPaths` in `kody-rules/_components/item.tsx` renders `**/*{.js,.jsx,.ts,.tsx}` as `....jsx, .ts, .tsx}`).
- **Precedence:** `path` is treated as explicit author intent and is checked before `fileScope`.
- **Extensionless files:** `extensionScopeAppliesToFile` lets files like `Gemfile`, `Dockerfile` and `Rakefile` through, while a `**/*{.rb,...}` glob excludes them.

Would persisting the mapping as `fileScope` fit better here? Happy to hear if there's a reason to prefer `path`.

Issue #1886, opened 2026-09-09
AWS WAF blocks Kody Rules writes whose examples contain XSS-like snippets (403 before the API)


malinosqui:
## Summary

`POST /api/proxy/api/kody-rules/add-library-kody-rules` returns **403** for some library rules. The block happens at the prod ALB's AWS WAF, before the request reaches Next.js or the API. The UI shows nothing (the library modal's `useAsyncAction` has no catch/toast). Surfaced by a customer support ticket on 2026-09-08 (org `MarketingInfra`, `79979761-1bf2-4bcc-84e6-3251ffa1f33b`).

## Evidence

ALB access logs (`s3://kodus-prod-alb-access-logs`, 2026-09-07 21:57–22:09Z), customer IP, 11 attempts:

```
POST https://app.kodus.io:443/api/proxy/api/kody-rules/add-library-kody-rules  elb=403  target=-  actions_executed=waf  matched_rule=-1
```

Every other request from the same client in the same minutes is `waf,forward` and succeeds (including other POSTs). The API log group has no matching `[403]`, confirming the request never arrived.

## Reproduction

Replaying all 500 library rules as anonymous POSTs through the proxy (expected answer is a 401 from the API):

- 498 → `401` (reached the API)
- 2 → `403` HTML page from the WAF:
  - `08d4da8d-d35f-4901-85c7-23284097e004` "Always sanitize user inputs" (the one the customer was adding)
  - `33c45a80-cbfa-4b16-a90f-074b81e06bb3` "Prevent Memory Allocation Vulnerabilities"

Bisecting the body: `title` alone → 401, `rule` alone → 401, **`examples` alone → 403**. The first rule's examples contain `<script>alert("Hacked!")</script>`, which matches the WAF's XSS body inspection. The same check will block `create-or-update` for any customer who writes a Kody Rule with an XSS/SQLi-looking example.

## Fix

Infra (needs WAF console access, the CLI IAM user has no `wafv2:*`):

- Exempt the Kody Rules write paths from the XSS/SQLi body rules (scope-down or a path-based allow rule), on both hops served by `kodus-application-prod-v2`:
  - `app.kodus.io/api/proxy/api/kody-rules/*`
  - `api.kodus.io/kody-rules/*`
- Consider the same for `/cli/kody-rules/*` (CLI rule sync sends code snippets too).

Follow-up in the web app: surface the failure in the library modal instead of swallowing it, so the next occurrence is visible.

PR #1915, opened 2026-09-14
fix(web): surface import failures + forward rule language (#1886)

andyst-dev:
[Issue #1886](https://github.com/kodustech/kodus-ai/issues/1886) | [Issue #1832](https://github.com/kodustech/kodus-ai/issues/1832) | [Backend half: PR #1914](https://github.com/kodustech/kodus-ai/pull/1914)

## Why the change

Importing a rule from the Kody Rules library failed without any message when the request was rejected (for example by the WAF), and it dropped the rule's language, so a rule meant for one language ended up applying to every file.

## Special things to note

- On its own, this PR only makes the API accept `language`. The import use case still ignores it, so rules stay unscoped until #1914 lands. Both PRs add the same optional DTO field, so whichever merges second rebases cleanly.
- The DTO field is required, not optional cleanup: the global `ValidationPipe` runs with `forbidNonWhitelisted: true`, so sending `language` without it turns every import into `400 property language should not exist`.
- The toast only shows the API message when `response.data.message` is a string. A WAF HTML page, a network error, or a validation 400 (where `message` is an array) all show the generic "blocked before reaching us" text. The WAF 403 itself still needs an exemption in the AWS console.

## Change outline

The import request body gains an optional field. It is only sent when the library rule has a language.

```diff
 POST /kody-rules/add-library-kody-rules   (AddLibraryKodyRulesDto)
 {
   title, rule, severity, examples, origin, status, scope,
   path: "",                        // still empty unless importing into a directory
+  language?: string,               // @IsOptional @IsString, e.g. "ruby"
   repositoriesIds, directoriesInfo, teamId
 }
```

The modal's import action now catches a failed request and shows an error toast instead of ending silently.

```diff
 addToRepositories()                       (library/kody-rules/@modal/[id]/_components/modal.tsx)
   newRule = { ...library rule fields, path: "" }
+  if rule.language: newRule.language = rule.language
   if directoryId: newRule.path = "<directory>/**"
-  addedKodyRules = await addKodyRuleToRepositories(...)
+  try
+    addedKodyRules = await addKodyRuleToRepositories(...)
+  catch error
+    message = error.response.data.message   // only if it is a string
+    toast destructive "Could not add rule \"<title>\"" (message ?? generic WAF hint)
+    return
   if centralized PR response: toast proposal, return
   reset kody-rules queries, toast success, navigate
```

Once #1914 is merged, the `language` field lets the backend give the rule a path.

```text
modal ──language──▶ AddLibraryKodyRulesDto (this PR: accepted)
                      └─▶ AddLibraryKodyRulesUseCase ──▶ resolveLibraryRulePath (#1914: empty path → language glob)
```


andyst-dev:
You are right, and this one would have broken every import, not just the failing ones. Fixed in `79334b5`: `language` is now a declared field on `AddLibraryKodyRulesDto` (`@IsOptional()` + `@IsString()`), so `ValidationPipe({ whitelist: true, forbidNonWhitelisted: true })` no longer answers `400 property language should not exist`.

I took the second option you listed rather than dropping the field, because the field is the whole point of the forwarding: the backend uses it to derive the path glob (#1832), otherwise the rule is persisted with an empty path and applies to every file. The DTO field is the same optional string as the one added in #1914 (I diffed both branches: byte-identical), which is where the derivation lives. Whichever of the two lands first, the other rebases without a delta.

One thing worth stating explicitly: until the backend half is in, the API validates and ignores `language`, so the rule is stored with its empty path exactly as it is today. No behavioural regression, and imports stop 400-ing either way.

Checked the rest of the request body against the DTO while I was in there: the modal sends `title`, `rule`, `path`, `severity`, `examples`, `origin`, `status`, `scope` and the service adds `repositoriesIds`/`directoriesInfo`/`teamId`. All are declared, so nothing else trips `forbidNonWhitelisted`.

apps/web/src/app/(app)/library/kody-rules/@modal/[id]/_components/modal.tsx
```diff
@@ -136,12 +142,44 @@ export const KodyRuleLibraryItemModal = ({
...
+                    typeof detail === "object" &&
+                    detail !== null &&
+                    "message" in detail &&
+                    typeof (detail as { message?: unknown }).message ===
+                        "string"
+                        ? (detail as { message: string }).message
+                        : undefined;
+
+                toast({
+                    variant: "destructive",
```
Comment (malinosqui) [inline apps/web/src/app/(app)/library/kody-rules/@modal/[id]/_components/modal.tsx:175]: Our toast doesn't have a `destructive` variant, so this renders without error styling.

```suggestion
                    variant: "danger",
```

apps/web/src/app/(app)/library/kody-rules/@modal/[id]/_components/modal.tsx
```diff
@@ -136,12 +142,44 @@ export const KodyRuleLibraryItemModal = ({
...
+                    title: `Could not add rule "${rule.title}"`,
+                    description:
+                        message ??
+                        "The request was blocked before reaching us (e.g. a security-layer rejection). Try again, or import the rule without the flagged example.",
```
Comment (malinosqui) [inline apps/web/src/app/(app)/library/kody-rules/@modal/[id]/_components/modal.tsx:179]: Only use the security-filter wording for an actual 403. Network errors and validation 400s also land here.

```suggestion
                    description:
                        message ??
                        ((error as { response?: { status?: number } })?.response
                            ?.status === 403
                            ? "The request was blocked by a security filter before reaching Kodus. Please try again later or contact support."
                            : "Something went wrong while adding this rule. Please try again."),
```

malinosqui (changes requested):
Thanks, surfacing the failure is exactly what #1886 needed. Two small fixes before we merge, both as one-click suggestions below:

1. **Toast variant:** `"destructive"` isn't one of our toast variants (`success | info | alert | warning | danger`, see `apps/web/src/core/components/ui/toaster/toast.tsx`), so this toast renders without the error styling. Other error toasts in the app use `"danger"`.
2. **Fallback message:** the WAF-specific text shows for every error that has no string `message`: network errors, and validation 400s whose `message` is an array. "Import the rule without the flagged example" also isn't something this modal lets the user do. The suggestion keeps the security-filter wording for 403s only.

The `language` change only takes effect together with the backend in #1914, so we'll merge the two together once both are ready.

PR #1917, opened 2026-09-14
fix(sso): read public API URL from API_URL instead of WEB_HOSTNAME_API

### LINKED ISSUE #1903: Self-hosted SAML broken: web app uses in-cluster API hostname as public SSO URL
## Problem

The web app builds the browser-facing public API URL from `WEB_HOSTNAME_API` instead of `API_URL`.

- `WEB_HOSTNAME_API` is the in-cluster Service address for the web pod's server-side `/api/proxy/api/*` calls: [values.yaml](https://github.com/kodustech/kodus-installer/blob/main/charts/kodus/values.yaml#L67-L68)
- `API_URL` is the public, browser-reachable API origin the API already uses for SAML callbacks: [api-url.util.ts](https://github.com/kodustech/kodus-ai/blob/main/libs/ee/sso/utils/api-url.util.ts)

`getApiPublicUrl()` in [apps/web/src/core/utils/api-public-url.ts](https://github.com/kodustech/kodus-ai/blob/main/apps/web/src/core/utils/api-public-url.ts) currently ignores `API_URL`, so the sign-in page sends the browser to `https://kodus-api/auth/sso/login/<organizationId>`, which only resolves inside the cluster.

cursedcoder:
## Problem

`getApiPublicUrl()` builds the browser-facing SSO URL from `WEB_HOSTNAME_API`/`WEB_PORT_API`, which is documented (in the self-hosted chart's `values.yaml`) as the **in-cluster** address the web pod uses for its own server-side `/api/proxy/*` calls. That address isn't reachable from the user's browser.

On a self-hosted deployment behind a split-horizon DNS setup (public hostname resolves externally, internal service name resolves in-cluster), this sends the browser to something like:

```
https://kodus-api/auth/sso/login/<organizationId>
```

which never resolves outside the cluster, breaking self-hosted SAML/SSO login. Confirmed on a live deployment — the sign-in page's injected `window.__KODUS_PUBLIC_CONFIG__.apiPublicUrl` was literally `"https://kodus-api"` (the bare in-cluster service name).

`libs/ee/sso/utils/api-url.util.ts` already documents `API_URL` as the single source of truth for the public API origin — used for the SAML callback — and its own comment says it's "the same env the frontend exposes as `apiPublicUrl`". This PR makes `getApiPublicUrl()` actually do that, instead of reusing `WEB_HOSTNAME_API`.

## Fix

- `getApiPublicUrl()` (server branch) now reads `process.env.API_URL` directly, matching `buildApiUrl()` in `libs/ee/sso/utils/api-url.util.ts`.
- Removed the now-unused `createUrl` import.
- `WEB_HOSTNAME_API` is untouched and still used everywhere it should be (the web pod's internal proxy).

No chart/values changes needed — `API_URL` is already part of the shared config every service (including `web`) receives.

Fixes #1903

## Test plan

- [ ] Deploy with distinct `WEB_HOSTNAME_API` (in-cluster) and `API_URL` (public) values and confirm the sign-in page's injected config now shows `apiPublicUrl` equal to `API_URL`, and that SSO login redirects to the public origin.
- [ ] Confirm the web pod's own server-side `/api/proxy/*` calls are unaffected (still use `WEB_HOSTNAME_API`).


cursedcoder:
Duplicate of #1913, which was opened first and also adds test coverage. Closing in favor of that one.

PR #1921, opened 2026-09-15
fix(telemetry): report self-hosted release version in heartbeats

### LINKED ISSUE #1827: Self-hosted telemetry reports package.json version (2.1.8) instead of the release version
## Bug

Self-hosted telemetry heartbeats report `kodus.version` from `package.json` (`2.1.8`, unchanged since April 2026) instead of the release the instance is actually running (`selfhosted-2.1.3x`).

As a result the whole fleet looks identical in the beacon: on 2026-09-01, 182 active instances report `2.1.8` and 18 report `2.1.7`, while the newest self-hosted tag is `selfhosted-2.1.33`. We cannot tell which instances have upgraded, which still lack features shipped in August (for example the self-hosted trial request CTA from `selfhosted-2.1.32`), or how fast releases roll out.

## Where

- `libs/telemetry/application/services/heartbeat-collector.service.ts:259` — `readKodusVersion()` reads `package.json` → `version` and returns it as `kodus.version`.
- The real release version already exists in the container: `RELEASE_VERSION` is passed as a build arg by `.github/workflows/selfhosted-build-push.yml` (line 393) into `docker/Dockerfile` (`ARG`/`ENV RELEASE_VERSION`), and `apps/api/src/services/version-check.service.ts:48` already uses `process.env.RELEASE_VERSION` for the version-status endpoint.

## Expected

`kodus.version` in the heartbeat payload matches the self-hosted release tag (e.g. `2.1.33`), normalized the same way `VersionCheckService` does (strip `selfhosted-` / `v` prefixes). Fall back to `package.json` only when `RELEASE_VERSION` is unset (`local` builds).

## Repro

```sql
-- kodus_telemetry (Neon)
select last_version, count(*) from telemetry_instances
where last_seen_at >= now() - interval '30 days' group by 1;
-- 2.1.8 | 182
-- 2.1.7 | 18
```

## Notes

- `package.json` `version` has not been bumped by the release process, so it is not a substitute for the tag.
- Consider also adding the raw tag (or `RELEASE_VERSION` as-is) to the payload so pre-release builds (`-rc.N`) stay distinguishable.

RainLib:
## Summary

Fixes https://github.com/kodustech/kodus-ai/issues/1827.

Self-hosted heartbeat telemetry now prefers the image's `RELEASE_VERSION` over the stale package version. It trims whitespace and removes the `selfhosted-` or `v` prefix while preserving prerelease/build suffixes. Unset, blank, and `local` releases retain the cached `package.json` fallback.

## Changelog routing (driven by the PR title prefix)

`fix(telemetry)` routes this change to Bug fixes.

## Test plan

- [x] Collector, self-hosted beacon, and beacon HTTP unit suites: 3 suites, 46 tests passed, including 11 new release-version cases.
- [x] ESLint on both changed files.
- [x] Environment schema drift and coverage checks.
- [x] `git diff --check`.
- [ ] Full suite / deployed heartbeat verification (not run).

## Notes for the reviewer

The change is isolated to version reporting; no payload fields, telemetry opt-in behavior, database migrations, or release pipeline changes are introduced. Existing package-file caching/failure fallback is unchanged. Prerelease identifiers remain in the existing version field, so no raw-tag field/schema change is necessary.

Local Jest's standard TypeScript config loader fails on the existing top-level `require` / `import.meta` combination (TS2441/TS1343). The three suites were run using the same repository configuration loaded in memory, removing only the `createRequire` declaration and converting its default export to CommonJS. No Jest configuration changes are included. The pre-push hook was bypassed because it invokes that broken loader; targeted tests and environment checks were run separately.

## Acceptance criteria

- Release images report RELEASE_VERSION rather than a stale package version.
- Strip selfhosted- or v prefixes and whitespace while retaining prerelease/build suffixes.
- Unset, blank, and local releases use the existing package.json fallback; no telemetry fields or opt-in behavior change.


Wellington01:
@kody start-review

PR #1923, opened 2026-09-15
docs(github): complete app permissions and connection checks

### LINKED ISSUE #1856: Docs: GitHub App setup guide omits required permissions & webhook-domain guidance, leading to silent non-functional installs
I got stuck connecting my self-hosted Kodus instance to GitHub via a GitHub App, and it took several hours — even working through it with AI assistance — before we found the actual cause. The app *looked* installed on both GitHub and Kodus's side, webhooks were delivering with `200 OK`, and there was still no error anywhere to point at. What follows is what AI (Claude) proposes as documentation updates to `how_to_deploy/en/platforms/github/github_app.mdx`, based on exactly what went wrong in that session.

(Filing here since `kodustech/docs` has issues disabled. Also possibly related: #1695 — the missing E2E coverage on the GitHub App path may be exactly why this never got caught.)

---

## What actually went wrong

1. **The guide never mentions setting Repository permissions on the App.** Creating a GitHub App with every permission left at "No access" is a *valid* GitHub configuration — it installs fine, shows "Installed", but with "Repository access: No repositories" and nothing to configure. Kodus never surfaces this as an error; it just never becomes "connected". This was the actual root cause, and nothing in the doc (or the linked GIFs) says to configure permissions at all.
2. **The Webhook URL step is ambiguous between `WEB_DOMAIN` and `API_DOMAIN`.** The two URLs immediately above it in the guide both use `WEB_DOMAIN`, so it's the natural (wrong) thing to reuse for the webhook. Pointing the webhook at `WEB_DOMAIN` doesn't error — GitHub gets a `200`/`307` back from the web app's own routing/auth-guard, and the delivery looks "successful" in Recent Deliveries, but the payload never reaches the `webhooks` service and no integration is ever created. This cost real debugging time because everything *looked* fine from GitHub's side.
3. **No way to verify the integration actually linked.** A delivered `installation`/`installation_repositories` webhook feels like confirmation, but it isn't — the Kodus-side integration record is only created by the browser completing the **Setup URL** round trip. We only found this by checking the Postgres `auth_integrations` table directly.

## Suggested patch to `how_to_deploy/en/platforms/github/github_app.mdx`

Ordered to match GitHub's own app-creation form (Callback/Setup → Webhook → Permissions → Events → Credentials):

**1. Strengthen the existing "Redirect on update" note** (Step 2) — it matters not just for the first install, but every time permissions are edited on an already-installed app:

> Check the option: **Redirect on update**
> This ensures that after installing the app, users are redirected back to Kodus to complete setup — **and this also fires the next time you edit the app's permissions on an already-installed instance.** If it's unchecked, approving new permissions or adding repositories won't relink the installation in Kodus even though GitHub shows it as configured.

**2. Fix the Webhook URL ambiguity** (Step 3):

> - **Webhook URL**:
>   ```
>   WEBHOOK_DOMAIN/github/webhook
>   ```
>   > ⚠️ For most self-hosted single-VM setups, `WEBHOOK_DOMAIN` **is your `API_DOMAIN`**, not `WEB_DOMAIN` — even though `WEB_DOMAIN` is used for the two URLs directly above. Pointing this at `WEB_DOMAIN` doesn't error; the delivery looks "successful" in Recent Deliveries, but the payload never reaches Kodus and no integration is ever created.

**3. New step — Configure Permissions** (currently missing entirely):

> ## Configure Permissions
>
> Under **Repository permissions**, grant:
>
> | Permission | Access |
> |---|---|
> | Contents | Read and write |
> | Issues | Read and write |
> | Pull requests | Read and write |
> | Checks | Read and write |
> | Metadata | Read-only *(added automatically)* |
>
> > ⚠️ Leaving every permission at "No access" is a valid GitHub App configuration. It installs and shows as "Installed", but with "Repository access: No repositories" and nothing to configure — and Kodus will never see it as connected. If the app looks installed on GitHub but Kodus never shows a connected repo, check this first.

**4. New step — Subscribe to events** (cross-reference the existing PAT-webhook doc, which already lists these but isn't linked from here):

> ## Subscribe to Webhook Events
>
> Kodus consumes exactly three events — selecting extras just sends traffic that's discarded:
> - Pull request
> - Pull request review comment
> - Issue comment
>
> *(Same three events as the [manual webhook setup](/how_to_deploy/en/platforms/github/github_webhook) — this App-based flow uses the same handler.)*

**5. Small note on the credentials step** — GitHub's own App page nudges users toward a Client ID ("Using your App ID to get installation tokens? You can now use your Client ID instead"), which doesn't apply here and can end up in `.env` for no reason:

> > Kodus's GitHub App integration doesn't use the Client ID — only `API_GITHUB_APP_ID`, `API_GITHUB_CLIENT_SECRET`, and `API_GITHUB_PRIVATE_KEY`. The Client ID is only relevant for the separate [OAuth "Sign in with GitHub"](/how_to_deploy/en/platforms/github/github_oauth) setup.

**6. New closing section — verifying it actually worked:**

> ## Verifying the connection
>
> A `200` on a webhook delivery (even for `installation` or `installation_repositories`) does **not** mean Kodus registered the integration — those events keep things in sync but don't create the link themselves. The link is only created when your browser completes the round trip through the **Setup URL** after installing/reconfiguring the app.
>
> After installing, confirm in the Kodus UI (**Git Settings**) that a repository actually shows as connected. If it doesn't:
> - Re-open the install URL (or **Configure** at `github.com/settings/installations`) and make sure you click through to a final **Save** on the repository-access screen — it's easy to land on that screen and leave without saving.
> - Confirm **Redirect on update** is checked (Step 2).
> - Confirm the Webhook URL is your **API** domain, not your web domain (Step 3).

---

Happy to turn this into a PR against `kodustech/docs` if someone can re-enable issues/PRs there, or point me at the right place.

RainLib:
## Summary

Fixes https://github.com/kodustech/kodus-ai/issues/1856.

Complete all five localized GitHub App setup guides with repository permissions, event subscriptions, installation approval and browser setup completion, and an end-to-end connection checklist. Explain why 200/307 responses alone do not prove webhook processing.

## Test plan

- [x] Parsed all five changed pages with Prettier's MDX parser and checked the setup checklist coverage.
- [x] Checked event names against GithubController: pull_request, pull_request_review_comment, issue_comment, and push.
- [x] Checked integration creation and GitHub service API usage; linked official GitHub registration documentation.
- [x] git diff --check.
- [ ] Fresh GitHub App installation and deployed docs rendering (not performed).

## Notes for the reviewer

The issue's proposed three-event list is stale: the current controller also accepts push. Contents is read-only for review access; write access is documented separately for file-committing features to avoid unnecessary permissions. API_DOMAIN is valid only with an explicit proxy route to the webhooks service. Ignored installation webhooks do not create the integration.

Docs-only change. Pre-push is bypassed because the existing local Jest config loader has a require/import.meta conflict; no runtime changes are included.

## Acceptance criteria

- All five locales document repository permissions, approval of installation permission changes, and the supported pull_request, pull_request_review_comment, issue_comment, and push events.
- Webhook URLs reach the webhooks service; API-domain usage requires a matching proxy route, and HTTP 200/307 alone is not proof of processing.
- Users finish the Setup URL browser flow and verify the intended repository is connected plus a test PR reaches webhook/worker processing.
- Contents write permission is distinguished from read-only review access; runtime integration behavior is unchanged.


Wellington01:
@kody start-review

PR #1926, opened 2026-09-15
fix(git): preserve host when rotating token (#1874)

### LINKED ISSUE #1874: Allow rotating a GitLab PAT without resetting the integration
### Is your proposal related to a problem?

Self-hosted Kodus connected to GitLab via PAT. Token rotation is a normal ops task (expiry, leak, switching to a dedicated bot account).

Once GitLab is connected, Settings → Git only shows **Reset**. There is no “update token”.

Reset is a disconnect, not a credential swap:

- deletes GitLab project webhooks (`DeleteIntegrationUseCase` → `deleteWebhook`)
- deletes `auth_integrations` and integration configs (including selected repositories)
- “Reset integration and remove repositories config” also wipes code-review parameters / PR messages / some Kody Rules

After that you re-select repos and wait for webhooks to be recreated. That is too much blast radius for “the PAT expired”.

Related: #1723 (first-connect modal traps the user; changing token required breaking out to Reset).

### Describe the solution you'd like

On the connected Git provider card, add **Update token** (GitLab first; same gap exists for GitHub/Bitbucket/Azure/Forgejo PAT).

Reuse the existing API. `POST /code-management/auth-integration` already overwrites credentials in place:

- `GitlabService.authenticateWithToken` → `handleIntegration`
- if an integration exists → `updateAuthIntegration` writes encrypted `authDetails.accessToken`

Keep selected repositories and existing webhooks. Validate the new PAT (`GET /api/v4/user` + list projects) before saving, same as first connect.

### Describe alternatives you've considered

1. **Reset + reconnect in the UI** — works, but drops webhooks and repo selection.
2. **Call `POST /code-management/auth-integration` by hand** — this already updates the token; it is not exposed in the UI after connect.
3. **Patch `auth_integrations.authDetails` in Postgres** — token is encrypted; easy to get wrong.

(1) is the only documented UI path. (2) is the real fix; the UI just needs to offer it.

### Additional context

Self-hosted Docker, GitLab (self-managed or gitlab.com), token auth (`authMode: "token"`).

Relevant code:

- `apps/web/src/app/(app)/settings/git/_components/connected-provider.tsx` — Reset only
- `apps/web/src/app/(app)/settings/git/_components/_modals/reset-integration-modal.tsx`
- `libs/platform/infrastructure/adapters/services/gitlab.service.ts` — `handleIntegration` / `updateAuthIntegration`
- `apps/api/src/controllers/codeManagement.controller.ts` — `POST /code-management/auth-integration`

### Requirement/Document

https://docs.kodus.io/how_to_use/en/code_review/general_config/gitlab_pat

---

##### House rules

- If this issue has a `🚨 needs approval` label, don't start coding yet. Wait until a core member approves feature request by removing this label, then you can start coding.
    - For clarity: Non-core member issues automatically get the `🚨 needs approval` label.
    - Your feature ideas are invaluable to us! However, they undergo review to ensure alignment with the product's direction.

andyst-dev:
## Summary

Fixes #1874

Rotating a Git provider PAT currently requires a full **Reset**, which disconnects webhooks, clears repository selection, and removes code-review configuration. This adds an **Update token** action that reuses the existing credential-upsert endpoint and preserves the integration.

### Change

- Add **Update token** for GitLab, Bitbucket, Azure Repos, and Forgejo. GitHub remains excluded because its connected flow is OAuth-app based.
- Reuse `POST /code-management/auth-integration`, including its existing token validation and in-place credential update.
- Return the stored non-secret host in connection status for GitLab, Bitbucket Data Center, and Forgejo.
- Prefill the modal from that host so rotating a token preserves the self-hosted instance instead of falling back to a public endpoint.
- Invalidate the integration query and revalidate `/settings/git` after success; surface a 400 as `Invalid Token`.

### Verification

- Host-preservation regression suite: **3 passing tests**, RED→GREEN verified against the pre-fix code for GitLab, Bitbucket Data Center, and Forgejo.
- Related targeted service suites: **38 passing tests**.
- Targeted ESLint: clean.
- New spec Prettier check: clean.
- TypeScript syntax transpilation: clean for all changed files.

Web-layer typecheck remains unavailable in this worktree because `apps/web/node_modules` dependencies such as `lucide-react` and `@tanstack/react-query` are absent. The modal reuses the existing connect/reset component signatures.


andyst-dev:
Thanks — fixed in `0b0213417`.

`/integration/connections` now exposes the existing non-secret instance host for GitLab, Bitbucket Data Center, and Forgejo. The Update token modal receives it, initializes its self-hosted toggle and URL from it, and therefore validates and persists rotations against the current instance unless the user deliberately changes that setting. The PR title is now `git settings: preserve host when rotating a token (#1874)` (57 characters).

Validation: exact frontend and backend files transpile cleanly; backend ESLint has 0 errors (only 7 pre-existing warnings in these large service files).

andyst-dev:
Follow-up validation in `8a6cd85ac` adds regression coverage for all three self-hosted providers. The tests assert that GitLab, Bitbucket Data Center, and Forgejo expose the stored host while no token or password enters the connection-status payload.

RED→GREEN is verified against the pre-fix code (all three cases fail without the host mapping). Targeted result: 38 passing tests; ESLint and Prettier checks are clean for the new spec.

The previous full backend CI failure was unrelated to this PR: 904 suites passed and the only failing integration suite could not find the Postgres `organizations` / `parameters` tables. This push has triggered a fresh CI run.

andyst-dev:
Thanks — two of these need a factual answer, and the rest are in place.

**"The three new tests pass with or without the change" (09-16 07:28)** — I checked this empirically rather than assuming: with the three service files restored to their pre-fix state (`git merge-base` = `9826f739`), the suite goes **3 failed / 3 passed**. The assertions read `status.config.host`, which the pre-fix `verifyConnection` does not return on these paths, so they are not satisfied by pre-existing behaviour.

The underlying concern was fair, though: nothing covered the rotation itself, only the backend host exposure. Added `update-token-integration-modal.spec.tsx` (2 tests) asserting that the payload handed to `createCodeManagementIntegration` still carries the stored host, plus the cloud case where it stays `undefined`. RED→GREEN verified by neutralising `host` in the modal: the self-hosted case fails with `host: undefined` while the cloud case stays green. Both suites together: **5 passed**.

**Bitbucket Cloud / Data Center resolution (09-16 07:01)** — `BitbucketService.getImplementation()` returns `bitbucketDataCenterService` only when `authDetails.host` is non-empty; otherwise it returns `bitbucketCloudService`. A Cloud connection carries no host, so the Data Center adapter cannot serve it, and the host it reads in its own connection status can only be a Data Center connection's host.

**Title (09-15 05:56)** — now `fix(git): preserve host when rotating token (#1874)`, 51 characters.

**Modal host prefill (09-15 05:56)** — implemented: `connected-provider.tsx` passes `connection.config.host`, the modal seeds `selfhosted` / `selfHostedUrl` from it, and sends it back on rotation.

**Eval evidence (09-16 07:01)** — added to the first 1000 characters of the description, including the not-run note for the end-to-end browser flow.

PR #1927, opened 2026-09-15
fix(pipeline): conclude check when an unexpected error escapes the stage loop (#1849)

### LINKED ISSUE #1849: GitHub check stays IN_PROGRESS forever when a review run fails internally — check must be concluded (failed/neutral)
## Problem

When a code-review run fails internally (e.g. the configured LLM provider is unreachable), the GitHub status check stays `IN_PROGRESS` indefinitely instead of being concluded with a `failure` (or `neutral`) result. Downstream automation (merge gates, branch protection) blocks on the hanging check, and there is no signal to the PR author that kodus failed.

## Observed behavior

Self-hosted kodus (`openai_compatible` BYOK provider, no fallback connectivity):

- The review agent dispatches with the configured provider.
- The provider call fails at the connection level (timeout, unreachable endpoint) with no HTTP status/body captured (see #1829 — the error details are dropped from logs).
- The automation execution lands in `partial_error` / a subsequent run is stuck `in_progress` in the DB with an empty `errorMessage`.
- The `byok.llm_errors_threshold` notification fires.
- The GitHub check remains `IN_PROGRESS` with no conclusion — it never flips to failed/neutral/cancelled.

## Expected behavior

If a review run fails internally (provider outage, worker error, pipeline error), kodus should **conclude the check** via the Checks API with `conclusion: failure` (or `neutral` for provider-side errors) and a short message. Leaving it `IN_PROGRESS` forever makes kodus the silent blocker.

## Suggested fix

- In the completion handler (where `partial_error`/`permanent`/mismatch states are resolved), always post a check conclusion for runs that terminated without a successful review — including internal/provider failures, not just explicit user actions.
- Treat "worker died / job stranded" (see #1830) and "internal stage failed" the same way: the check must not outlive the job.

## Related

- #1829 — BYOK provider errors drop statusCode/responseBody, obscuring the root cause
- #1830 — workflow jobs stranded in PROCESSING when a worker dies mid-job
- #1844 — PR summary path has no BYOK fallback; main-provider outage ends run in partial_error

andyst-dev:
## Summary

Fixes #1849

A code-review run that failed **outside** a stage's own try/catch never concluded its GitHub check: `PipelineExecutor.execute()` only called `onPipelineFinish` at the end of the stage loop, so an unexpected throw between stages (skip/jump logic, an observer leaking during `onStageStart`, etc.) escaped the function entirely. `AutomationCodeReviewService` then marked the execution `ERROR`, but there was no observer call to finalize the check — leaving the platform check `IN_PROGRESS` indefinitely (branch protection / merge gates block forever, PR authors get no signal).

### Change

Wrap the stage loop in a try/catch guard. On an unexpected escape, force `statusInfo.status = ERROR`, append a `critical` pipeline error, and let the existing `onPipelineFinish` path do its job — posting a `failure` conclusion so the check no longer outlives the job. The catch does not rethrow: the pipeline is already broken past that point, and the run must be recorded as failed rather than blowing up.

- `pipeline-executor.service.ts`: try/catch around the stage loop; on escape → ERROR status + critical error + `onPipelineFinish` still fires (observer concludes the check in failure).
- `pipeline-executor.service.spec.ts`: regression test — a throw between stages (mocked `handleSkipOrJump`) now forces `ERROR` and still fires `onPipelineFinish`.

### Verification

- RED→GREEN checked: stashing the fix makes the new test fail (`1 failed`); with the fix, the spec suite passes (14 tests).
- `tsc --noEmit -p tsconfig.lib.json`: clean for this file. `eslint`: 0 errors/warnings.

Note: `pipeline-executor.service.ts` does not currently pass stock Prettier (pre-existing — the loop body uses a non-standard indentation that Prettier would reformat across the whole file), so this file was intentionally left un-reformatted to keep the PR a minimal, reviewable diff.


andyst-dev:
Thanks — fixed in `b89a7e593`. The `logger.error` in the unexpected-error catch now carries `organizationAndTeamData: (context as any)?.organizationAndTeamData ?? null` in its metadata, alongside `...context?.pipelineMetadata`, `correlationId` and `status`, so an aborted pipeline stays traceable to its tenant like every other logger call in this file.

Spec still green (14 tests).

malinosqui (changes requested):
Thanks for picking up #1849! The guard itself is harmless, but after tracing the code on `main` I don't think this is the path that leaves the check `IN_PROGRESS`, so I'd like to confirm the root cause before we merge it as the fix.

### The failure in #1849 is already caught

Links point at `main` @ `58cca21`.

- **Stage errors:** `stage.execute()`, `notifyStageCompletion()` and the metrics/logging after them all run inside the per-stage `try/catch` ([pipeline-executor.service.ts#L179-L285](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/core/infrastructure/pipeline/services/pipeline-executor.service.ts#L179-L285)). A provider that is unreachable throws from inside a stage, so the error lands there. It gets recorded as a `critical` error and the loop keeps going.
- **Observer errors:** every observer call (`onPipelineStart`, `onStageStart`, `onPipelineFinish`, ...) goes through `notifyObservers`, which catches and logs errors itself ([L361-L378](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/core/infrastructure/pipeline/services/pipeline-executor.service.ts#L361-L378)). An observer that throws can't escape the loop.
- **Skip/jump logic:** `handleSkipOrJump` only logs and calls `produce`. No production stage calls `skipToStage` today, so that branch doesn't run in practice.
- **Check conclusion:** after the loop, `onPipelineFinish` runs. `CodeReviewPipelineObserver` concludes the check as `FAILURE` whenever the status is `ERROR` or any error is `critical` ([code-review-pipeline.observer.ts#L170-L186](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/code-review/infrastructure/observers/code-review-pipeline.observer.ts#L170-L186)).

So with a provider outage, `main` should already conclude the check as a failure. The new test mocks the private `handleSkipOrJump` to throw. That shows the `catch` works, but not that this path is reachable in a real run.

### More likely causes

Any of these would leave the check `IN_PROGRESS`, and a `try/catch` around the loop would not help with any of them:

1. **The worker dies or hangs mid-job** (#1830). Neither the stage `catch` nor `onPipelineFinish` ever runs.
2. **`finalizeCheck` itself fails.** The Checks API call is wrapped in a `try/catch` that only logs `Failed to finalize check` ([pipeline-checks.service.ts#L477-L483](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/core/infrastructure/pipeline/services/pipeline-checks.service.ts#L477-L483)). The check stays open.
3. **There's no `checkRunId` in the observer context when the pipeline finishes.** `finalizeCheck` then returns silently (`No checkRunId found in context for finalizeCheck`, logged at debug level).
4. **A stage returns `undefined` instead of a context.** `notifyStageCompletion` reads `context.statusInfo` and throws inside the per-stage `try`. The per-stage `catch` then reads `context.statusInfo` again while logging ([L249](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/core/infrastructure/pipeline/services/pipeline-executor.service.ts#L249)), and that second throw escapes the loop. That is a real escape, but this PR doesn't cover it either: the new `catch` runs `produce(context, ...)` with `context === undefined` and throws again.

### How to check what's actually happening

Please try to reproduce #1849 before changing the executor:

1. Run the stack locally or self-hosted, with a BYOK `openai_compatible` provider pointed at an endpoint nothing listens on (e.g. `http://127.0.0.1:9/v1`), and no fallback model.
2. Open a PR so a review runs, then watch the check on the PR and the worker logs.
3. In the logs for that run (filter by `correlationId`), look for:
   - `Stage '<name>' failed:`: the provider error was caught by the stage.
   - `Finished pipeline: ...`: the loop finished and `onPipelineFinish` was called.
   - `Failed to finalize check` or `Observer onPipelineFinish failed`: the conclusion was attempted and failed (cause 2).
   - `No checkRunId found in context for finalizeCheck`: cause 3.
   - None of the above after the stage start logs: the worker died or hung (cause 1, #1830).
4. Repeat once and kill the worker container in the middle of the review (`docker kill <worker>`), to compare with the worker-death case.

If one of these runs shows an error escaping the stage loop (a stack trace with no `Stage '...' failed` and no `Finished pipeline`), please paste it here and add a test that reproduces that real path, without mocking a private method. That would make this PR the right fix.

### Requested changes

- **Show the real path:** share reproduction or log evidence that this path is reached. If there isn't any, change `Fixes #1849` to `Refs #1849`, so the issue doesn't close while the check can still hang.
- **Clear `skippedReason` in the new `catch`.** Otherwise the "restore skipped status" block right after the loop ([L288-L297](https://github.com/kodustech/kodus-ai/blob/58cca21e539e825587719628000ce49bdb1ad14b/libs/core/infrastructure/pipeline/services/pipeline-executor.service.ts#L288-L297)) overwrites `ERROR` with the skipped status, and the check concludes as `SKIPPED` instead of `FAILURE`.
- **`undefined` context (cause 4):** either handle it (for example, keep the last valid context before calling `produce`) or leave it out of the PR's claims.

PR #1928, opened 2026-09-15
fix(platformData): UUID suggestion IDs for ingest (#1846)

### LINKED ISSUE #1846: Fine-tuning never indexes persisted suggestions: ingest requires UUID ids but persistence writes Mongo ObjectIds
Found a bug? Please fill out the sections below. 👍

### Issue Summary

The fine-tuning pipeline **cannot index suggestions that were actually persisted**. Suggestion subdocument ids are generated as **MongoDB ObjectId hex strings** (24-char), but both the Mongo read path and the embed insert path **require RFC UUID** ids. Everything that is stored is dropped before embeddings are created, so implemented suggestions and 👍/👎 never become learning examples.

Id assignment (persistence):

- `PullRequestsRepository.newSubDocumentId()` → `new mongoose.Types.ObjectId().toString()`
- `addSuggestionToFile` / `addFileToPullRequest` use the same ObjectId string when `id` is missing

Id filtering (learning ingest):

- `findByOrganizationAndRepositoryWithStatusAndSyncedFlag` `$filter`s suggestions with `$regexMatch` against a UUID regex, so ObjectId ids never leave Mongo.
- `SuggestionEmbeddedService.isValidSuggestion` requires `UUID_REGEX.test(s.id)` before `bulkCreateFromMongoData`.
- The unit test `rejects an id that is a string but not a UUID` encodes this mismatch as intended behaviour.

`KodyFineTuningService.getImplementedSuggestions` / feedback sync therefore see an empty candidate set even when closed PRs have `implementationStatus: implemented` and reactions.

### Steps to Reproduce

1. Run a code review that posts file-level suggestions (default persist path, no pre-assigned UUID on the suggestion).
2. Confirm in Mongo that `files.suggestions.id` is a 24-character hex ObjectId (not `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`).
3. Implement or 👍 a suggestion so it is eligible for `getDataForEmbedSuggestions` (closed PR, `implemented` or non-neutral feedback, `improvedCode` present).
4. Trigger `KodyFineTuningStage` (`startAnalysis` → `syncronizeSuggestions` → `bulkCreateFromMongoData`).
5. Observe `suggestion_embedded` stays empty; the aggregation returns suggestions arrays stripped of ObjectId ids.

Any other relevant information: this is independent of whether OpenAI embeddings are configured. Even with a valid embedding key, ingest never sees the stored rows.

### Actual Results

- Persisted suggestion ids are Mongo ObjectIds.
- Fine-tuning load + `isValidSuggestion` keep only UUID-shaped ids.
- Learning corpus is not populated from real reviews.

### Expected Results

- Ids the repository writes are ids the fine-tuning ingest accepts (ObjectId hex **or** UUID).
- `findByOrganizationAndRepositoryWithStatusAndSyncedFlag` does not `$regexMatch` away every stored suggestion.
- `isValidSuggestion` still rejects empty/missing ids, but accepts the format `newSubDocumentId()` produces.
- Implemented / reacted suggestions on closed PRs are eligible to be embedded (subject to existing label filters, `improvedCode`, etc.).

### Technical details

- `libs/platformData/infrastructure/adapters/repositories/pullRequests.repository.ts` — `newSubDocumentId()` (~1647), `addSuggestionToFile` (~1660), aggregation UUID `$regexMatch` in `findByOrganizationAndRepositoryWithStatusAndSyncedFlag` (~1192–1244). There is a second copy of the same UUID filter later in the same file (~1309).
- `libs/kodyFineTuning/infrastructure/adapters/services/suggestionEmbedded/suggestionEmbedded.service.ts` — `UUID_REGEX` + `isValidSuggestion`; `bulkCreateFromMongoData` filters with it first and returns `[]` when none pass.
- `libs/kodyFineTuning/infrastructure/adapters/services/suggestionEmbedded/suggestionEmbedded.service.spec.ts` — `isValidSuggestion` `rejects an id that is a string but not a UUID`.
- Related: `KodyFineTuningService.syncronizeSuggestions` can still mark PRs `syncedEmbeddedSuggestions: true` when `suggestionsToEmbed` is empty (PRs are returned even if nothing was embeddable). After this id bug is fixed, consider only flipping the sync flag when something was actually inserted.

### Evidence

Code inspection + existing unit test: persistence generates ObjectIds; ingest requires UUIDs; the spec asserts non-UUID strings are rejected. A regression test should use a 24-char hex id (what `new mongoose.Types.ObjectId().toString()` returns) and expect `isValidSuggestion` / the aggregation to keep the suggestion, not drop it.

andyst-dev:
## Summary

Fixes #1846

Suggestion ids were generated as Mongo ObjectId hex strings, but both the Mongo read path (`findByOrganizationAndRepositoryWithStatusAndSyncedFlag`) and the embed ingest (`SuggestionEmbeddedService.isValidSuggestion`) filter by an RFC UUID regex. New suggestions were therefore dropped before embedding.

### Change

- Keep `newSubDocumentId()` ObjectId-shaped for `files[].id`.
- Add `newSuggestionId()` for UUID-shaped suggestion ids.
- Use the suggestion generator in the bulk file-change flow and in `addSuggestionToFile`'s repository fallback.
- Add regression coverage for both id formats and their distinct bulk-flow call sites.

Existing rows with ObjectId suggestion ids are not migrated in this scoped fix; newly written suggestions now meet the ingest contract.

### Verification

- Targeted repository and bulk-file-change suites: **32 passing tests**.
- RED→GREEN verified for both the bulk-flow generator split and the `addSuggestionToFile` UUID fallback.
- Targeted ESLint: clean.
- TypeScript syntax transpilation: clean for all changed files.


andyst-dev:
Thanks — addressed in `9c527f4bd`.

The identifier generators are now explicit: file sub-documents keep ObjectId-shaped ids, while suggestions use UUIDs only. The service bulk path and the repository fallback both use the suggestion generator, and regression coverage verifies the two paths independently.

The title is now `platformData: generate UUID suggestion IDs for ingest (#1846)` (61 characters).

Validated with the targeted repository and bulk-file-change suites: 31 passing tests; targeted ESLint is clean.

andyst-dev:
Follow-up validation in `023a07957` covers the repository fallback directly: when `addSuggestionToFile` receives no id, it must call `newSuggestionId()` and persist the generated UUID.

The new assertion is RED→GREEN verified against the ObjectId fallback and the targeted repository + bulk-flow suites now pass 32 tests.

The previous full backend CI failure was unrelated to this PR and identical to #1926: the integration runner could not find the Postgres `organizations` / `parameters` tables. This push has triggered a fresh CI run.

andyst-dev:
Both findings are fixed in `61628f189`.

**1. `newSuggestionId()` missing on the service (TS2420).** Correct: `IPullRequestsService extends IPullRequestsRepository`, so adding the member to the repository interface also binds `PullRequestsService`. It now delegates to the repository exactly like `newSubDocumentId`:

```ts
newSuggestionId(): string {
    return this.pullRequestsRepository.newSuggestionId();
}
```

Verified with the repo's own typecheck (`tsc --noEmit --skipLibCheck`): `pullRequests.service.ts(36,14): error TS2420` was reported before the change and is gone after, with no other error on the two touched files.

**2. Bare global `crypto`.** Correct as well — the module relied on `globalThis.crypto`, which only resolves on Node >= 19. Added the explicit `import * as crypto from 'crypto';` at the top of the repository, matching how the rest of the codebase imports it.

The 50 tests in `pullRequests.repository.spec.ts` and `pullRequests.service.bulk-file-changes.spec.ts` still pass.

PR #1970, opened 2026-09-18
chore(evals): delete scripts nothing runs any more

malinosqui:
## Eval evidence
- Proof: not a quality change — no eval runs differently. Deleted files are unreachable from every entry point (scripts, workflows, promptfoo configs, docs, `require` graph), so recall is untouched: `sets.light` stays at 0.24 under `gpt-5.6-luna`.
- Not run: nothing to run. `node evals/engine-gate.js --profile=harness`, `jest evals/` (98 passed) and `pnpm eval:wiring` (460 files loaded, all covered) are green on this branch.

Stacked on #1951 — review that one first; this targets its branch.

`evals/` carries 92 tracked `.js`; 18 of them are reached by nothing at all. This deletes those.

- **15 from the kody-rules sharding investigation** (`sharded-experiment`, `sharded-v2`, `shard-cost-repro`, `scope-fanout-repro`, the polyglot harvesters, the detector feasibility probes). The question they answered shipped as file × rule sharding; the scripts are history, and git keeps history.
- **`compare-recall.js`** was the old PR gate (base vs head with a noise tolerance), replaced by the nightly floor + confirmation. Its entry in the preflight's required files goes with it.
- **`guard-pairs-runner.js`**, **`capture-return.js`**: no caller.

Not touched here, because it needs a call rather than a delete: the promptfoo path (`investigation/run-eval.js`, three `promptfoo*.yaml`, 8 `eval:investigation*`/`eval:trace-context` scripts) is superseded by `run-recall.js`, but 66 dataset entries carry `assert: file://investigation-assertion.js`, so retiring it means either leaving those dangling or rewriting the datasets. Happy to stack that next.


Wellington01:
Two leftovers from the delete, both in files the diff doesn't touch:

`evals/kody-rules/pull-prod-rules.js:90` closes with `Now run: node evals/kody-rules/backfill-shipped.js --model=gpt-5.4-mini`, and that file goes away here. The surviving next step is the one `detectors-1831.json` already documents: `detector-fp-repro.js --rules=prod-rules`. Line 3 of the same header names both `backfill-shipped.js` and `prod-rules-judge-eval.js`, so it needs the same pass.

`evals/kody-rules/behavioral-scoring.ts` loses both importers in this PR (`behavioral-shipped.js:62` and `run-convention-cases.js:47`), leaving `behavioral-scoring.spec.ts` as the only caller — a 55-line lib and a 74-line spec that nothing else reaches. Same criterion as the rest of the PR, just `.ts` instead of `.js`.

Wellington01 (approved):

